PR #3, opened 2021-01-29
Task 2 create student model

josh-cooper:
## Description
Create a student model and factory to generate students.
Rspec is also configured with factory bot here but Rsepc gem and specs are not required until task 3.

## Breakdown of Commits
Configure rspec : 7bb27b9a2d8689244355b72e3bf568bd697aff42

Add factory bot dependency : 0f3fa1eb3250ac6350997f48ebeee030a4e225b0

Generate student scaffolding : 0612e2107de6cae96281606e8406e65af11c9f5f

Migrate Student : b141b77526d641e70c08d57392692f0bf0eb6f44

Add Student validations : 7798e44ca16d2a2462039d47c84f0641b7220d06

Fix missing test-unit dependency : dde02906e1255f84486b5700345ed0f469c66522

Capitalize titles in student factory : c20612aa77e6bb1bbed2bca6d2b5fa7a0dc05356

Cleanup : 9e3b6c4c0240d7bacb238941c1f5af88bc855c41
 - Move dependencies into the new development and test groups

#### TODO Checklist
* [x] Create a Student model with title, first_name, middle_name, last_name, email, birth_date, gender
* [x] add factory_bot_rails gem and display some example usage in the PR description

  
#### Factory bot example usage
```
irb(main):001:0> FactoryBot.build(:student)
=> #<Student id: nil, title: "Miss", first_name: "John", middle_name: "Alan", last_name: "Smith", email: "user1@example.com", birth_date: "2011-01-28 05:17:01", gender: "male", created_at: nil, updated_at: nil>
irb(main):002:0> FactoryBot.build(:student)
=> #<Student id: nil, title: "Mr", first_name: "John", middle_name: "Alan", last_name: "Smith", email: "user2@example.com", birth_date: "2011-01-28 05:17:01", gender: "female", created_at: nil, updated_at: nil>
irb(main):003:0> FactoryBot.build(:student)
=> #<Student id: nil, title: "Mr", first_name: "John", middle_name: "Alan", last_name: "Smith", email: "user3@example.com", birth_date: "2011-01-28 05:17:01", gender: "male", created_at: nil, updated_at: nil>
```


my-sms/spec/spec_helper.rb
```diff
@@ -0,0 +1,96 @@
...
+  # Print the 10 slowest examples and example groups at the
+  # end of the spec run, to help surface which specs are running
+  # particularly slow.
+  config.profile_examples = 10
+
+  # Run specs in random order to surface order dependencies. If you find an
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  config.order = :random
```
SyntheticDave, line 88: It's not a requirement for this project, but I strongly recommend enabling this in general, as leaving it disabled can hide order dependent spec failures.

my-sms/spec/spec_helper.rb
```diff
@@ -0,0 +1,96 @@
...
+
+  # Seed global randomization in this process using the `--seed` CLI option.
+  # Setting this allows you to use `--seed` to deterministically reproduce
+  # test failures related to randomization by passing the same `--seed` value
+  # as the one that triggered the failure.
+  Kernel.srand config.seed
```
SyntheticDave, line 94: Enabling this is really useful when the above is enabled, as you can reproduce the 'random' order that caused the spec failure.

my-sms/app/assets/stylesheets/scaffolds.css.scss
```diff
@@ -0,0 +1,73 @@
+body {
```
SyntheticDave, line 1: You may end up changing/removing this when it comes to task 4, which includes styling with bootstrap. I have no issues with leaving it here for now though.

my-sms/app/helpers/students_helper.rb
```diff
@@ -0,0 +1,2 @@
+module StudentsHelper
+end
```
SyntheticDave, line 2: I know this comes from the scaffold, and is all good to be here, but I thought I'd note that we often add view level logic to models using decorators or presenters.
https://github.com/drapergem/draper is a common decorator gem, which allows you to specify at the controller level that a resource or collection should be decorated for the view. 
Not something for this initial project, but a common design pattern that we use, which allows us to reduce our reliance on helpers.

my-sms/app/assets/javascripts/students.js.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
SyntheticDave, line 1: Not sure if this project needs any JS. You can delete this if not.

my-sms/Gemfile
```diff
@@ -19,6 +19,7 @@ end
 group :development, :test do
   gem 'rspec-rails', '~> 3.6.0'
   gem 'factory_bot_rails', '~> 4.10.0'
+  gem 'test-unit'
```
SyntheticDave, line 18: I didn't realise this was an actual dependency of rails 3.2, so my bad for suggesting you remove it in the first place.
They've removed it as a dependency in subsequent versions.

SyntheticDave (approved):
Looking good! 
I've left some comments, but no actual requests to change anything.
There are some files generated by the scaffold that you'll want to remove or modify in the upcoming tasks (e.g. you'll need all tests passing from task 3), but I don't see any issue with leaving them as-is for now. 
I suggest cleaning up/removing some of the files you don't need in step task 4, when you'll be looking at view code.

my-sms/spec/spec_helper.rb
josh-cooper: Sounds good I'll add this on.

my-sms/Gemfile
josh-cooper: Haha all good that was Matt's you commented on actually.

my-sms/app/helpers/students_helper.rb
josh-cooper: That's pretty interesting I might actually update my task 3 to include a decorator since currently the full_name method just exists on the model.

my-sms/Gemfile
SyntheticDave: Oh, man, I'm getting slightly mixed up in my reviews 😅

my-sms/app/helpers/students_helper.rb
SyntheticDave: It really depends on whether that method will be needed on the model outside of the views (as by convention we only decorate records with draper decorators for the views).
There's no problem putting `full_name` on the model itself, it's a good default place for a method like this, so you might still want to do so anyway, but decorators are a great way to extract view-only logic from your models.

PR #4, opened 2021-02-01
Task 3 add full name to student

josh-cooper:
## Description
Add full_name method to the Student model.
Add specs for Student, including the full_name method.
Move full_name method to a decorator and remove helpers scaffolding.

## Breakdown of Commits
Add student model specs : d4783bf8c22e05bdcebeb5827d5ef9317556e9d7

Update bundle : 9d8a80edb715409d9c1c7455c95db9f34a8e1539

Add rubocop config from JR RTO : b5539b4b4ab36dc3dc0c91ec0732baaf390dd3f6

Exclude rubocop blocklength linting on certain files : a727054ef91a33d4a0eecf2582a304a50b44d3c5

Add additional student models specs : fe1d0d8ca8dd4f6a9401a31dc202e1818886e041

Fix rubocop config : 90143fcc997727fd31058d33a7763ceb8eb53b5f
    
 - Update TrailingEmptyLines cop name
 - Enable DisplayCopNames

Add additional student model specs : 7fa32c11f92427da53610550cef9c35ec66e8703
 - Test presence validation on email, birth date, and title
 - Test lack of presence validation on middle_name
 - Test email format validation

Fix rubocop config : ac51a831be9d293519a7a6719157685b9f8cf582
 - Match spec files independent of where rubocop is run

Add students request spec. : 6666fd3aa300c3ce6885dbe3bef84fcb3d0e24b4

Add student full name method : 60f4fa6b3e97ca9732b28207677a12531b142e23

Move full_name method to decorator : b3a75a6dfd15555dd285b08759ed52a3965f671e
 - Add draper dependency
 - Generate draper decorator for Student
 - Move full_name method to decorator
 - Move full_name model specs to student_decorator_spec
 - Add calls to decorate in Student controller

Add Students controller specs : d43a7ef9731f0878771f385820f809f5ed66deb7

Remove helpers scaffolding : 1fcc1caf7c81a2f982bf1a86e9364503accb9911
 - Remove empty application and students helpers
 - Remove students helper spec

Refactor controller specs: ac4705697ef9337a0a1ceb4b8abc83c4cbadf23e
 - Implement subject, let and context blocks to make test subject more explicit and improve DRYness.

Refactor Student controller GET specs: 2111138268fed04a87c2b4bfdc06289769bde868
 - Individual describes for each route
 - Change student object creation to use let method and define explicit subject for each route

Refactor Student controller DELETE #destroy spec: 551e65cb9e69cc4946a34aa25e79fee7598465dd
 - Extract student creation into let method

Refactor Student controller POST #create specs: d3409cf965894121842dd82db141a492a96175ca
 - Move factory methods out of subject into let methods inside each context

Refactor Student controller PUT #update specs: 670f46c3d489436ed9ca3318ae6d2eeee57e29a4
 - Move factory methods out of subject into let methods inside each context
 - Cleanup carriage return

Use shoulda matchers in Student controller specs: 6cccd5ae169559b61c06f9a81151c62838717fdc

Cleanup Student controller spec: 6cb157930c38c2517ba8f3705025ecbdf9e3adbd
 - Remove valid_session let
 - Remove unneccessary references to valid_session

Merge branch 'develop' into task_3_add_full_name_to_student: d2c866de5cc087173559ec7b36cd895fc6946cf0
 - Resolve merge conflicts

Refactor Student controller GET specs : 2111138268fed04a87c2b4bfdc06289769bde868
    
 - Individual describes for each route
 - Change student object creation to use let method and define explicit subject for each route

Refactor Student controller DELETE #destroy spec : 551e65cb9e69cc4946a34aa25e79fee7598465dd
    
 - Extract student creation into let method

Refactor Student controller POST #create specs : d3409cf965894121842dd82db141a492a96175ca
    
 - Move factory methods out of subject into let methods inside each context

Update my-sms/spec/controllers/students_controller_spec.rb : a70a6bf72bd766828081bbe7d4aae7b3b4799dda
    
    Co-authored-by: jhuizy <jhuizenga99@gmail.com>

Merge branch 'task_3_add_full_name_to_student' of https://github.com/josh-cooper/my-sms into task_3_add_full_name_to_student : 934e333e9afac005a0568aba6cde545221b82f6c

Refactor Student controller PUT #update specs : 670f46c3d489436ed9ca3318ae6d2eeee57e29a4
 - Move factory methods out of subject into let methods inside each context
 - Cleanup carriage return

Use shoulda matchers in Student controller specs : 6cccd5ae169559b61c06f9a81151c62838717fdc

Refactor student controller decoration : c622c9b79de51c35bb6c9a9eeac84a8dbe795faf
    
 - Refactor to use top level decorates_assigned

Refactor student retrieval into before filter : 1446af57446f508f529adc3c17b423d4c5da53c3

Clean up full_name method. : f070817eabf0bc4f6fc9ddecb6fcf48e1cc909a7
 - Replace string interpolation with array join.

Refactor specs to automatically include FactoryBot methods : a1a9ea184bcabc7ce5d1917ec246dbdae2e1f7e0

Refactor Student controller test. : 7b355989292cab5b4ebc1d4043e89b841347e435
    
 - Move assignment to let methods
 - Explicitly test for change on updates

Test new record is not created : d1c657a7ddbcc2e460c38ad78d1400a7799c5297

Refactor DELETE #destroy specs : b7da5ea777f1c6443f26dfaacc81841bf07d8888
    
 - Move student creation to before block to make change in Student count clearer

Reload student attributes when checking for change : 29ca75f6dc2e2a539c4c3f1b65e9de0404e8b54a

Refactor student_spec : f45e16ad5f92b912d1aeee870101c3a8026d406f
    
 - Add shoulda-matcher dependency
 - Introduce testing subject and contexts
 - Move student creation into let methods
 - Replace parameter validation with shoulda matchers

Cleanup : 33e35d29ea868341028d39ab9a3f6787de9d78f4
    
 - Remove leftover :focus tag on spec

#### TODO Checklist
* [x] Add a full_name method to the Student model (title first_name middle_name last_name)
* [x] Add rspec rails gem
* [x] Add spec for full_name method


my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  describe 'GET #index' do
+    it 'returns a success response' do
+      get :index, {}, valid_session
+      expect(response).to be_success
+    end
+  end
+
+  describe 'GET #show' do
+    it 'returns a success response' do
+      student = FactoryBot.create(:student)
```
jhuizy: You can use [let bindings](https://rubydoc.info/gems/rspec-core/RSpec/Core/MemoizedHelpers/ClassMethods#let-instance_method) to setup test variables to reuse between test cases

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+      get :show, { id: student.to_param }, valid_session
```
jhuizy: Could also use `subject` to make it explicit what you're testing and integrate with `should` matchers in the `it` blocks.

```ruby
subject { get :show, { id: ... } }
it { should be_success }
```

Or more explicitly

```ruby
subject {... }
it "does something" do
  expect(subject).to eq(10)
end

File: my-sms/spec/controllers/students_controller_spec.rb [hunk 3]
```diff
@@ -0,0 +1,136 @@
...
+  describe 'POST #create' do
+    context 'with valid params' do
+      it 'creates a new Student' do
+        expect do
+          post :create, { student: FactoryBot.attributes_for(:student) }, valid_session
+        end.to change(Student, :count).by(1)
+      end
+
+      it 'redirects to the created student' do
+        post :create, { student: FactoryBot.attributes_for(:student) }, valid_session
```
jhuizy: Here the `subject {...}` pattern would make your tests more DRY (not repeating code unnecessarily)

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  describe 'PUT #update' do
+    context 'with valid params' do
+      it 'updates the requested student' do
+        student = FactoryBot.create(:student)
+        new_attributes = FactoryBot.attributes_for(:student)
+        put :update, { id: student.to_param, student: new_attributes }, valid_session
+        student.reload
+
+        attrs = %i[first_name last_name gender]
+        student_attrs = student.attributes.fetch_values(*attrs.map(&:to_s))
```
jhuizy: You can use the `%w` macro to create a list of strings rather than converting symbols to strings

jhuizy:
The use of draper is interesting. The JR plus repo already has the dependency but doesn't look like it's used much. There are a few `presenter`s in the repo which basically do the same thing as draper `decorator`s but in a more manual way. I think you'll find that the philosophy of rails is to have "fat" models and so alot of the time things like `full_name` would just be defined within the model itself, but it's entirely opinionated I guess 😄

jhuizy:
Also, seems there's a few merge conflicts that need to be resolved first

jhuizy (changes requested):
Merge conflicts need resolving plus other issues mentioned can be looked at

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: Had an issue where the new_attrs used symbols and the student_attrs wanted strings.

josh-cooper:
> The use of draper is interesting. The JR plus repo already has the dependency but doesn't look like it's used much. There are a few `presenter`s in the repo which basically do the same thing as draper `decorator`s but in a more manual way. I think you'll find that the philosophy of rails is to have "fat" models and so alot of the time things like `full_name` would just be defined within the model itself, but it's entirely opinionated I guess 😄

Implemented draper after Dave Reece mentioned it in the last task. Wasn't necessary but wanted to try it out since it was used in the JR plus repo.

jhuizy:
> Implemented draper after Dave Reece mentioned it in the last task. Wasn't necessary but wanted to try it out since it was used in the JR plus repo.

Haha yea... I just got picked up for not using it in one of my PRs 😅

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I've added ac4705697ef9337a0a1ceb4b8abc83c4cbadf23e to refactor the tests a bit to use subjects and let bindings.

josh-cooper:
> Also, seems there's a few merge conflicts that need to be resolved first

Fixed

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,123 @@
...
-      student = Student.create! valid_attributes
-      get :index, {}, valid_session
-      expect(response).to be_success
+  describe 'GET' do
+    subject(:student) { FactoryBot.create(:student) }
+
+    context '#index' do
+      it 'returns a success response' do
+        get :index, {}, valid_session
+        expect(response).to be_success
```
jhuizy: Here I would put the student in a let binding, and wrap the `get :index` in a subject.

```suggestion
    let(:student) { FactoryBot.create(:student) }
    subject { get :index, {}, valid_session }

    context '#index' do
      it 'returns a success response' do
        expect(subject).to be_success
```

You can then take this one step further by using `shoulda` matchers instead of `expectation` matchers, like:

```ruby
describe "GET /student/:id" do

  # Subject is always the same in this `describe` block
  subject { get :show, { id: student.id } }

  context "with valid student" do

    # with each context we can change the value of `student`
    # which will in turn update our request since it uses the `student.id`
    let(:student) { create :student }
    
    it "does the right thing" do
      # using the shoulda matchers implicitly calls the subject so
      # there is no need to manually call it
      should be_success
      should render_template('student/show')
      # etc..
    end
  end
end
```

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,114 @@
...
-      expect(response).to redirect_to(students_url)
+    it 'destroys the requested student' do
+      expect { subject }.to change(Student, :count).by(0)
     end
-  end
 
+    it 'redirects to the students list' do
+      expect(subject).to redirect_to(students_url)
+    end
+  end
```
jhuizy: This is a good example of `subject` here, but the `student = ...` should be moved to a let binding instead

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,114 @@
...
 
-    context "with invalid params" do
+    context 'with invalid params' do
+      let(:new_student_args) { { first_name: nil } }
+
       it "returns a success response (i.e. to display the 'edit' template)" do
-        student = Student.create! valid_attributes
-        put :update, {:id => student.to_param, :student => invalid_attributes}, valid_session
-        expect(response).to be_success
+        expect(subject).to be_success
```
jhuizy: Nice 👍

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,114 @@
...
+
+    context 'with valid params' do
+      it 'updates the requested student' do
+        subject do
+          student.reload
+          attrs = %i[first_name last_name gender]
+          student_attrs = student.attributes.fetch_values(*attrs.map(&:to_s))
+          new_attrs = new_attributes.fetch_values(*attrs)
+          expect(student_attrs).to eq new_attrs
+        end
```
jhuizy: This doesn't need to be wrapped in the subject here, but you will need to call the subject 

```suggestion
        subject
        student.reload
        attrs = %i[first_name last_name gender]
        student_attrs = student.attributes.fetch_values(*attrs.map(&:to_s))
        new_attrs = new_attributes.fetch_values(*attrs)
        expect(student_attrs).to eq new_attrs

```

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,114 @@
...
-      it "updates the requested student" do
-        student = Student.create! valid_attributes
-        put :update, {:id => student.to_param, :student => new_attributes}, valid_session
-        student.reload
-        skip("Add assertions for updated state")
+  describe 'PUT #update' do
+    let(:new_student_args) { {} }
+    subject do
+      student = FactoryBot.create(:student)
+      new_attributes = FactoryBot.attributes_for(:student, **new_student_args)
```
jhuizy: Put these both in let bindings

jhuizy:
Left a few more comments, hopefully it gives you some idea on how to structure the tests, otherwise let me know, happy to re-look at things or discuss further

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I've added this in 6cccd5ae169559b61c06f9a81151c62838717fdc

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I've fixed this pattern in 551e65cb9e69cc4946a34aa25e79fee7598465dd

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I've fixed this in 670f46c3d489436ed9ca3318ae6d2eeee57e29a4
Also applied it to d3409cf965894121842dd82db141a492a96175ca

josh-cooper:
> Left a few more comments, hopefully it gives you some idea on how to structure the tests, otherwise let me know, happy to re-look at things or discuss further

Thanks. Updated the PR again addressing those.

my-sms/app/controllers/students_controller.rb
```diff
@@ -1,8 +1,10 @@
+# frozen_string_literal: false
+
```
jhuizy, line 2: This is the default option (you would use `frozen_string_literal: true` to override), unless you're setting `RUBYOPT=--enable-frozen-string-literal`?

josh-cooper: I have some rubocop autoformatting and this was cleaner than disabling the cop.

jhuizy: Ah, all good 👍

jhuizy (approved):
LGTM

Not sure if one of the others wants to take a quick look over to review things according to any company standards I'm not familiar with or aware of yet

.rubocop.yml
```diff
@@ -0,0 +1,31 @@
+Metrics/AbcSize:
```
SyntheticDave, line 1: Great to see this here ✨

my-sms/app/controllers/students_controller.rb
```diff
@@ -1,8 +1,10 @@
...
 class StudentsController < ApplicationController
   # GET /students
   # GET /students.json
   def index
-    @students = Student.all
+    @students = StudentDecorator.decorate_collection(Student.all)
```
SyntheticDave: You should be able to remove your manual `decorate` statements with the following at the top of your controller:

```ruby
decorates_assigned :students
decorates_assigned :student
```

This will take any instance variable `@students` or `@student` in this case, and apply the appropriate decorator, which it infers based on the class name. I believe this will correctly handle the collection decoration in the case of `:students` (but let me know if you run into issues).

If you need to use a specific decorator that Draper can't infer from the record class name, you can use:

```ruby
decorates_assigned :students, with AnotherDecorator
```

my-sms/app/controllers/students_controller.rb
```diff
@@ -13,7 +15,7 @@ def index
   # GET /students/1
   # GET /students/1.json
   def show
-    @student = Student.find(params[:id])
+    @student = Student.find(params[:id]).decorate
```
SyntheticDave: You've got a few instances of `@student = Student.find(params[:id])` in this controller, which I'd recommend moving to a [before_filter](https://guides.rubyonrails.org/v3.2/action_controller_overview.html#filters) to DRY it up.

```ruby
# Alternatively, you could use only: [:show, :edit, :update, :destroy]. I just generally go with the shorter list
before_filter :find_student, except: [:index, :new, :create]
```

And then create a private method at the bottom of the controller:

```ruby
def find_student
  @student = Student.find(params[:id])
end
```

The `before_filter` will run before any of the methods specified, so you can just use `@student` in those methods if needed, as well as the relevant views.

my-sms/app/decorators/student_decorator.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class StudentDecorator < Draper::Decorator
+  delegate_all
+
+  def full_name
+    "#{title + ' ' if title.present?}"\
+    "#{first_name} "\
+    "#{middle_name + ' ' if middle_name.present?}"\
+    "#{last_name}"
```
SyntheticDave: I love that you're embracing string interpolation, but I find squeezing logic into those is pretty messy. What do you think of this possible alternative implementation?

```ruby
def full_name
  [title, first_name, middle_name, last_name].compact.join(' ')
end
```

Here, compact will clear any `nil` values, and the remaining are joined by a space. Note that compact only clears `nil`, rather than `blank` values, so empty strings will remain. If we need to account for the possibility of empty strings, we can do the following:

```ruby
def full_name
  [title, first_name, middle_name, last_name].select(&:present?).join(' ')
end
```

Just in case you haven't come across ruby's Symbol to Proc, these two statements are equivalent:

```ruby
collection.select(&:present?)
# Above is just a handy shortcut for below
collection.select { |value| value.present? }
```

You can use symbol to proc any time you just need to call a single method on whatever you're passing to the block. The syntax is a little odd, but often ends up being a little more readable.

Some additional explanation here if you're keen on some extra reading: https://blog.pjam.me/posts/ruby-symbol-to-proc-the-short-version/

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
-  # StudentsController. Be sure to keep this updated too.
-  let(:valid_session) { {} }
-
-  describe "GET #index" do
-    it "returns a success response" do
-      student = Student.create! valid_attributes
-      get :index, {}, valid_session
-      expect(response).to be_success
+RSpec.describe StudentsController, type: :controller do
+  let(:student) { FactoryBot.create(:student) }
```
SyntheticDave: Minor one, but you can add some config to your rails helper to automatically include the FactoryBot methods: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#rspec

After which, this line becomes:

```ruby
let(:student) { create(:student) }
```

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
-  describe "GET #show" do
-    it "returns a success response" do
-      student = Student.create! valid_attributes
-      get :show, {:id => student.to_param}, valid_session
-      expect(response).to be_success
+  describe 'GET #show' do
+    subject { get :show, { id: student.to_param } }
+    it 'returns a success response' do
+      should be_success
+      should render_template('show')
```
SyntheticDave, line 43: We usually look to keep to a single assertion per expectation/it block, but this is a good example of where it makes more sense to group both into the one test. Good work.

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
+      let(:new_attributes) { FactoryBot.attributes_for(:student) }
 
-      it "updates the requested student" do
-        student = Student.create! valid_attributes
-        put :update, {:id => student.to_param, :student => new_attributes}, valid_session
+      it 'updates the requested student' do
+        subject
         student.reload
-        skip("Add assertions for updated state")
+        attrs = %i[first_name last_name gender]
```
SyntheticDave: So I think this test works, but I think we could clean it up a little. We generally try to do assignment outside of the test,  which helps keep the actual test blocks smaller and more contained to act, and assert, with any arranging happening outside that block.

Another potential issue is that as you're not using something like Faker to randomise your student's attributes, you're not actually guaranteeing that your update works, as the attributes you're passing in, so I would recommend setting the values either in the initial record, or the update params to ones that are explicitly not the default for this factory.

What do you think of the following:

```ruby
# Set the initial values to whatever you want, as long as they don't match the factory defaults

let!(:student) { FactoryBot.create(:student, first_name: 'foo', last_name: 'bar', gender: 'baz') }
let(:new_attributes) { FactoryBot.attributes_for(:student).stringify_keys }
let(:attr_names) { %w[first_name last_name gender] }
# Reload may or may not be necessary, recommend experimenting with that
let(:student_attrs) { student.reload.attributes.fetch_values(*attr_names) }
let(:new_attr_values) { new_attributes.fetch_values(*attr_names) }

it 'updates the requested student' do
  subject
  student.reload
  expect(student_attrs).to eq new_attr_values
end
```

For completeness, you could also consider the alternative assertion:

```ruby
it 'updates the requested student' do
  expect { subject }.to change { student_attrs }.to(new_attr_values)
end
```

Note that `expect` and `change` are passed a block in this case, so they can be evaluated at the correct times.

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
+        new_attrs = new_attributes.fetch_values(*attrs)
+        expect(student_attrs).to eq new_attrs
       end
 
-      it "redirects to the student" do
-        student = Student.create! valid_attributes
-        put :update, {:id => student.to_param, :student => valid_attributes}, valid_session
-        expect(response).to redirect_to(student)
+      it 'redirects to the student' do
+        should be_redirect
```
SyntheticDave, line 113: Probably don't need `should be_redirect` if you're following it with `should redirect_to(student)`, as the second assertion effectively encompasses the first. That being said, it also doesn't hurt to leave `should be_redirect` here, as it just checks the response code, so I'm not fussed either way.

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
+      let(:new_student_args) { { first_name: nil } }
+      let(:new_attributes) do
+        FactoryBot.attributes_for(:student, { first_name: nil })
+      end
+
       it "returns a success response (i.e. to display the 'edit' template)" do
-        student = Student.create! valid_attributes
-        put :update, {:id => student.to_param, :student => invalid_attributes}, valid_session
-        expect(response).to be_success
+        should be_success
```
SyntheticDave, line 125: Also consider testing that no new record is created:

```ruby
it 'does not create a new record'
  expect { subject }.to_not change { Student.count }
end
```

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -23,119 +25,115 @@
...
+    subject do
+      delete :destroy, { id: student.to_param }
     end
 
-    it "redirects to the students list" do
-      student = Student.create! valid_attributes
-      delete :destroy, {:id => student.to_param}, valid_session
-      expect(response).to redirect_to(students_url)
+    it 'destroys the requested student' do
+      expect { subject }.to change(Student, :count).by(0)
```
SyntheticDave: This _is_ testing that the delete works, as we know that a student record will be created when referenced. However, it's a misleading spec to read, as it appears that it's testing that the call has no effect.

What do you think of this instead?

```ruby
# Reference the student, so we know they're in the DB before the delete call
# You can do this in a before block, or in the first line of your `it` block.
before { student }

it 'destroys the requested student' do
  expect { subject }.to change(Student, :count).by(-1)
end
```

my-sms/spec/models/student_spec.rb
```diff
@@ -1,5 +1,66 @@
...
 
-RSpec.describe Student, :type => :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+RSpec.describe Student, type: :model do
+  it 'has a valid factory' do
+    expect(FactoryBot.build(:student)).to be_valid
+  end
+
+  it 'has a factory that generates unique emails' do
+    student = FactoryBot.build(:student)
```
SyntheticDave: As with the other specs, I'd recommend pulling the assignment out of the block to something like the following:

```
let(:student) { FactoryBot.build(:student) }
let(:other_student) { FactoryBot.build(:student) }

it 'has a factory that generates unique emails' do
  expect(student.email).not_to eq other_student.email
end
```

Also, I'm not suggesting you delete this, but we don't tend to test our factory behaviour. It's not a terrible idea though.

my-sms/spec/models/student_spec.rb
```diff
@@ -1,5 +1,66 @@
...
+    FactoryBot.create(:student, email: 'user@example.com')
+    student = FactoryBot.build(:student, email: 'user@example.com')
+    student.valid?
+    expect(student.errors[:email]).to include('has already been taken')
+  end
+
+  it 'is invalid without a first name' do
+    student = FactoryBot.build(:student, first_name: nil)
+    student.valid?
+    expect(student.errors[:first_name]).to include('can\'t be blank')
```
SyntheticDave: We generally prefer to use double quotes over single if that means you can avoid escaping apostrophes like this. It's good form to use single quotes whenever you're not using string interpolation, so I like that you've done that, but I would generally just use `"can't be blank"` in this case.

my-sms/spec/models/student_spec.rb
```diff
@@ -1,5 +1,66 @@
...
+
+  it 'is valid without a gender' do
+    expect(FactoryBot.build(:student, gender: nil)).to be_valid
+  end
+
+  it 'is valid without a middle name' do
+    expect(FactoryBot.build(:student, middle_name: nil)).to be_valid
+  end
+
+  it 'is invalid with an incorrect format' do
```
SyntheticDave: I would probably name this 'is invalid with an incorrect email format'

SyntheticDave (commented):
This is good work @josh-cooper!
I've left some chunky comments in here, but overall it's looking good, and gives you a good head-start into the next few tasks.
Hit me up for any clarifications you need, or disagreements you have 😉

my-sms/app/controllers/students_controller.rb
josh-cooper: I added this in 1446af57446f508f529adc3c17b423d4c5da53c3
Although opted for only over except.

my-sms/app/decorators/student_decorator.rb
josh-cooper: Thanks, join makes way more sense.
Came across Symbol to Proc in task 4 and I'm a big fan, I find the Ruby functional syntax a bit messy coming from arrow syntax in ES6/Kotlin/etc and this cleans it up a bit in simple cases.
Added your version in f070817eabf0bc4f6fc9ddecb6fcf48e1cc909a7

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: Added in a1a9ea184bcabc7ce5d1917ec246dbdae2e1f7e0

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I updated with your suggestions in 7b355989292cab5b4ebc1d4043e89b841347e435
I also experimented with reload and it turns out you don't need it. I replaced the block with the just the change expectation, however, I couldn't leave student_attrs in the let method as it will not re-evaluate the cache before and after the change. Therefore I've moved the attributes to check directly into the change block. I'm not sure if that's the best option here but it was cleaner and more idiomatic AFAIK than my other solution (putting a proc inside let and calling .call in the change block).

Thanks for stringify_keys, didn't know that existed.
Also the use of let! vs let was also something I hadn't considered.

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: I took this as following the same pattern as testing for a success code before testing which template is rendered, but I guess the relationship is reciprocal with a redirect.
I'll leave it in for now if its up to me as I'd prefer to keep the same pattern for the sake of using a pattern, even if it doesn't test any additional functionality.

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: Thanks, added in d1c657a7ddbcc2e460c38ad78d1400a7799c5297

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: Actually, this broke later on and I had to add reload back in. I guess if you aren't careful the ActiveModel objects aren't deterministic?

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: This is much clearer. Added in b7da5ea777f1c6443f26dfaacc81841bf07d8888

my-sms/app/controllers/students_controller.rb
josh-cooper: Thanks this is way cleaner now. 
Fixed in c622c9b79de51c35bb6c9a9eeac84a8dbe795faf

my-sms/spec/models/student_spec.rb
josh-cooper: Ended up installing shoulda-matchers as well and making the whole thing cleaner.
f45e16ad5f92b912d1aeee870101c3a8026d406f

my-sms/spec/models/student_spec.rb
josh-cooper: Understood. This ended up removed anyway in f45e16ad5f92b912d1aeee870101c3a8026d406f

my-sms/spec/controllers/students_controller_spec.rb
SyntheticDave: Yeah, the distinction between `let` and `let!` is a really important one, and useful to avoid needing to reference a variable just to create a record (which I might have suggested in another comment, oops).

Without the `reload`, the instance of the record is loaded into memory and rails doesn't know to reload from the database of this record is updated from your controller action, so the manual reload is necessary.

Because the controller is fetching the record separately, it's not making changes to the instance you've got in your test. You can see this more clearly by checking `student.object_id` in your `it` block vs the one fetched in the controller.

As an aside this is a really common gotcha when it comes to stubbing/mocking, where you'll add something like `allow(student).to receive(:first_name).and_return('foo')`, but because the instance in the controller is fetched separately, the stub isn't applied to the controller's `student`, so it returns the actual name on the record. In this case you need to either also stub the model's `find` to return your specific instance, or use `any_instance_of`.

I hope I've actually answered your question here. Let me know if I haven't.

my-sms/spec/controllers/students_controller_spec.rb
SyntheticDave: Yeah, I'm good with that. Being in the same block means it's not actually running the test twice, so I think it's reasonable to leave it there.

my-sms/spec/models/student_spec.rb
SyntheticDave: Great idea. I coulda mentioned that 👍

my-sms/spec/controllers/students_controller_spec.rb
josh-cooper: Thanks that's good to know.

my-sms/spec/controllers/students_controller_spec.rb
```diff
@@ -134,8 +134,10 @@
       delete :destroy, { id: student.to_param }
     end
 
+    before { student }
```
SyntheticDave, line 137: Small note, this could also be `let!(:student) { create(:student) }`, where the `let!` ensures this block is evaluated before your test. However, as you've got `let(:student) { create(:student) }` outside all the context blocks at the top of this test, I personally think this is a neater solution that's just as readable.

SyntheticDave (approved):
Great work! 👍

PR #5, opened 2021-02-02
Task 4 add student list page

josh-cooper:
## Description
Add a List page to display Student details showing full name with other available attributes.
Add pagination and bootstrap styling to list page.

## Breakdown of Commits
Change students list to display full name : 863be84c99a1a48ed4d59103abde333bb44b2c71

Add pagination to students listing : 349dd0a5a4547203c5a5e487627f376b09487ca3
 - Add will_paginate dependency
 - Change rendering all posts in students index to a paginated query
 - Render page buttons in students index erb
 - Set per page limit to 10

Add bootstrap gem and install static assets : 7b88b6a6920041fb8e7445a1cbb8affe082e66ab

Add bootstrap styling to Students : 05f500ab5658d33214e974ea5643eaa8187cd103
 - Generate default bootstrap layout for Student
 - Replace full name, pagination and show link in generated layout rather than adding in bootstrap styling manually
 - Use rails internationalisation helpers for students table headings

Override will_paginate link renderer : 216f2e040cd5e3a86e8890017a9213c39ad5a78b
    
    Add bootstrap styling to pagination links

Fix decorator interaction with pagination : 956af6b6057f2bbbb3a59d9573269a08169ff282
 - Define PaginatingDecorator to delegate pagination methods to will_paginate

Add custom CSS to Student model show route : 0c06baa6dbbdc5dcca34ad9c7c5302e6872862f9
 - Replace strong tags with spans and custom class
 - Add CSS file asset for default styles for model show routes displaying a list of attributes
  - Right align and bold attribute list headings, left align descriptions
 - Import new CSS asset in application CSS

Add AU I18n : b544ac29f98ce09c63899e275435d9cefcc11c40
 - Add standard AU I18n config file
 - Use long date format in Student model show route

Display flashes : 4f9c2f9e9c0210c5872efd072f6431d5efcf61fd
 - Use bootstrap styling to display flashes
 - Assign bootstrap alert class for particular flash classes
 - Create before_render callback
 - Add error messages to error flashes in before_render callback
 - CSS file for flash styles
 - Remove scaffolded error display

Move students per page default into constant : 39a3030580d87e5999959ca7d9f866b58f4e14e3

Refactor flash_errors to short circuit : 640d900caff9530e476b4857a1e4842cc82e001d

Add specs for error propagation from model to flash : 21acf9bd1752fa314e6f2e13f90c233c4e8a6543

Refactor internal method to access instance variable directly : 19b82cd3dc7de83283b7660a0c09c3b44eb187c8

Refactor erb files to use newer hash syntax : 5554e5f766237ceea0344d79b67ba1becda28022

#### TODO Checklist
* [x] Add a List page to display Student details showing full name with other available attributes.
* [x] Add pagination to the list page with will_paginate gem
* [x] Add css styling with Bootstrap Rails Gem

  
image

### DIFF
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -89,8 +89,9 @@ class StudentsController < ApplicationController
   end
 
   def flash_errors
+    return unless student&.errors&.full_messages
     student.errors.full_messages.each do |error|
       (flash[:error] ||= []) << error
-    end if student&.errors
+    end
   end
 end
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -9,7 +9,9 @@ class StudentsController < ApplicationController
   # GET /students
   # GET /students.json
   def index
-    @students = Student.paginate(page: params[:page], per_page: 30)
+    @students = StudentDecorator.decorate_collection(
+      Student.paginate(page: params[:page], per_page: 10)
+    )
 
     respond_to do |format|
       format.html # index.html.erb
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -9,7 +9,7 @@ class StudentsController < ApplicationController
   # GET /students
   # GET /students.json
   def index
-    @students = Student.all
+    @students = Student.paginate(page: params[:page], per_page: 30)
 
     respond_to do |format|
       format.html # index.html.erb
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -11,7 +11,7 @@ class StudentsController < ApplicationController
   # GET /students.json
   def index
     @students = StudentDecorator.decorate_collection(
-      Student.paginate(page: params[:page], per_page: 10)
+      Student.paginate(page: params[:page], per_page: Student::DEFAULT_PER_PAGE)
     )
 
     respond_to do |format|
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -5,6 +5,7 @@ class StudentsController < ApplicationController
   decorates_assigned :student
 
   before_filter :find_student, only: %i[show edit update destroy]
+  before_render :flash_errors, only: %i[edit create update]
 
   # GET /students
   # GET /students.json
@@ -85,5 +86,11 @@ class StudentsController < ApplicationController
 
   def find_student
     @student = Student.find(params[:id])
+  end
+
+  def flash_errors
+    student.errors.full_messages.each do |error|
+      (flash[:error] ||= []) << error
+    end if student&.errors
   end
 end
--- a/my-sms/app/controllers/students_controller.rb
+++ b/my-sms/app/controllers/students_controller.rb
@@ -89,8 +89,8 @@ class StudentsController < ApplicationController
   end
 
   def flash_errors
-    return unless student&.errors&.full_messages
-    student.errors.full_messages.each do |error|
+    return unless @student&.errors&.full_messages
+    @student.errors.full_messages.each do |error|
       (flash[:error] ||= []) << error
     end
   end
--- a/my-sms/app/helpers/application_helper.rb
+++ b/my-sms/app/helpers/application_helper.rb
@@ -13,6 +13,19 @@ module ApplicationHelper
     super *[collection_or_options, options].compact
   end
 
+  ALERT_CLASS = {
+    error: 'alert-danger',
+    success: 'alert-success',
+    alert: 'alert-warning',
+    notice: 'alert-info'
+  }.freeze
+  ALERT_CLASS_DEFAULT = 'alert-primary'
+
+  # auto assign bootstrap class depending on flash type
+  def get_alert_class(flash_type)
+    ALERT_CLASS.fetch(flash_type, ALERT_CLASS_DEFAULT)
+  end
+
   # custom link renderer to add bootstrap styling to pagination links
   class BootstrapLinkRenderer < WillPaginate::ActionView::LinkRenderer
     protected
--- a/my-sms/app/helpers/application_helper.rb
+++ b/my-sms/app/helpers/application_helper.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module ApplicationHelper
+  # override link renderer in will_paginate
+  def will_paginate(collection_or_options = nil, options = {})
+    if collection_or_options.is_a? Hash
+      options = collection_or_options
+      collection_or_options = nil
+    end
+    unless options[:renderer]
+      options = options.merge renderer: BootstrapLinkRenderer
+    end
+    super *[collection_or_options, options].compact
+  end
+
+  # custom link renderer to add bootstrap styling to pagination links
+  class BootstrapLinkRenderer < WillPaginate::ActionView::LinkRenderer
+    protected
+
+    def page_number(page) # rubocop:disable Metrics/MethodLength
+      current_number = page == current_page
+
+      aria_label = @template.will_paginate_translate(
+        :page_aria_label,
+        page: page.to_i
+      ) { "Page #{page}" }
+      page_link = link(
+        page,
+        page,
+        rel: rel_value(page),
+        class: 'page-link',
+        "aria-label": aria_label
+      )
+      tag(
+        :li,
+        page_link,
+        class: "page-item#{' active' if current_number}",
+        "aria-label": aria_label,
+        "aria-current": current_number ? 'page' : nil
+      )
+    end
+
+    # explicity assign pagination class to prevent breaking change
+    def html_container(html)
+      tag(:dive, html, container_attributes.merge(class: 'pagination'))
+    end
+  end
+end
--- a/my-sms/app/views/layouts/application.html.erb
+++ b/my-sms/app/views/layouts/application.html.erb
@@ -1,14 +1,31 @@
 <!DOCTYPE html>
 <html>
-<head>
-  <title>MySms</title>
-  <%= stylesheet_link_tag    "application", :media => "all" %>
-  <%= javascript_include_tag "application" %>
-  <%= csrf_meta_tags %>
-</head>
-<body>
-
-<%= yield %>
-
-</body>
+  <head>
+    <title>MySms</title>
+    <%= stylesheet_link_tag    "application", :media => "all" %>
+    <%= javascript_include_tag "application" %>
+    <%= csrf_meta_tags %>
+  </head>
+  <body>
+    <!-- display flashes -->
+    <% flash.each do |type, msg| %>
+      <div class="alert <%= get_alert_class(type) %>">
+        <!-- test if message is string or iterable -->
+        <% if msg.respond_to?(:each) %>
+          <ul class="alert-list">
+            <% msg.each do |msg| %>
+              <li class="alert-item"><%= msg %></li>
+            <% end %>
+          </ul>
+        <% else %>
+          <%= msg %>
+        <% end %>
+        <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+    <% end %>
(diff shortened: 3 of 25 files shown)


my-sms/app/views/students/show.html.erb
```diff
@@ -1,40 +1,31 @@
...
+  <dd><%= @student.middle_name %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:last_name) %>:</strong></dt>
+  <dd><%= @student.last_name %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:email) %>:</strong></dt>
+  <dd><%= @student.email %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:birth_date) %>:</strong></dt>
+  <dd><%= @student.birth_date %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:gender) %>:</strong></dt>
+  <dd><%= @student.gender %></dd>
+</dl>
```
jhuizy: I like the use of `dl`, `dt` and `dd` here for semantic html and accessibility. I think the `<strong>'s` could probably be replaced with a `<span class="your-class">` and then some css to style that appropriately (ie. make it bold).

my-sms/app/views/students/show.html.erb
```diff
@@ -1,40 +1,31 @@
...
+  <dt><strong><%= model_class.human_attribute_name(:first_name) %>:</strong></dt>
+  <dd><%= @student.first_name %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:middle_name) %>:</strong></dt>
+  <dd><%= @student.middle_name %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:last_name) %>:</strong></dt>
+  <dd><%= @student.last_name %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:email) %>:</strong></dt>
+  <dd><%= @student.email %></dd>
+  <dt><strong><%= model_class.human_attribute_name(:birth_date) %>:</strong></dt>
+  <dd><%= @student.birth_date %></dd>
```
jhuizy: Is the default date format nice and readable?

my-sms/app/views/students/_form.html.erb
```diff
@@ -1,45 +1,76 @@
...
+        <h3 class="panel-title"><%= pluralize(@student.errors.count, "error") %> prohibited this student from being saved:</h3>
+      </div>
+      <div class="panel-body">
+        <ul>
+        <% @student.errors.full_messages.each do |msg| %>
+          <li><%= msg %></li>
+        <% end %>
+        </ul>
+      </div>
     </div>
```
jhuizy: Another way to do this is to put it in a flash and define a way to display the flash on "standard" views/pages, making this error view a bit more re-usable

my-sms/app/helpers/application_helper.rb
```diff
@@ -0,0 +1,48 @@
...
+        "aria-current": current_number ? 'page' : nil
+      )
+    end
+
+    # explicity assign pagination class to prevent breaking change
+    def html_container(html)
+      tag(:dive, html, container_attributes.merge(class: 'pagination'))
+    end
+  end
+end
```
jhuizy, line 61: Did this code block get taken from somewhere? if so could be worth mentioning in a comment

jhuizy (commented):
Looks pretty good to me, just some minor discussion points

my-sms/app/helpers/application_helper.rb
josh-cooper: This is my own, although based it off some examples I saw.

my-sms/app/views/students/show.html.erb
josh-cooper: Added some custom classes in 7d8ac2e028d62a8a9cfb09bbec0907d51adc1c76

Didn't really end up changing much over the default bootstrap CSS applied to dl/dt/dd elements but was a lesson in the asset pipeline basics.

my-sms/app/views/students/show.html.erb
josh-cooper: Changed the format to something more nice and local via an AU I18n config.
b544ac29f98ce09c63899e275435d9cefcc11c40

my-sms/app/controllers/students_controller.rb
```diff
@@ -9,7 +9,9 @@ class StudentsController < ApplicationController
   # GET /students
   # GET /students.json
   def index
-    @students = Student.all
+    @students = StudentDecorator.decorate_collection(
+      Student.paginate(page: params[:page], per_page: 10)
```
jhuizy: As mentioned in Matt's PR, consider moving this magic number (`10`) to a constant or appropriately named variable (eg. `DEFAULT_PER_PAGE`)

my-sms/app/views/students/show.html.erb
jhuizy: Can't seem to see that commit, has it been pushed? (or maybe I'm blind 😄 )

my-sms/app/views/students/_form.html.erb
josh-cooper: Did this in 4f9c2f9e9c0210c5872efd072f6431d5efcf61fd
Defined a way to display flashes on standard views/pages and automatically created error flashes if there were errors on the Student model.

my-sms/app/views/students/show.html.erb
josh-cooper: Sorry I hadn't pushed 😅
Updated now

my-sms/app/controllers/students_controller.rb
josh-cooper: Fixed in 39a3030580d87e5999959ca7d9f866b58f4e14e3

my-sms/app/views/layouts/application.html.erb
```diff
@@ -1,14 +1,31 @@
...
+            <% end %>
+          </ul>
+        <% else %>
+          <%= msg %>
+        <% end %>
+        <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+          <span aria-hidden="true">&times;</span>
+        </button>
+      </div>
+    <% end %>
```
jhuizy, line 27: There was probably a bit of confusion from my last comment, I meant to only show the the fact that there _is_ an error in the flash, and then have more error details displayed inline. Consider replacing this with the `<%= bootstrap_flash %>` I mentioned in slack?

josh-cooper: I got that, just took it as an exercise to dive into what flashes could be used for and investigate the process the rails controller goes through.

I looked more into the bootstrap_flash helper to try and implement it now and it is slightly incompatible. My implementation handles an array of errors and I'd rather keep that implementation than try and put formatting directly into the flash server side.

my-sms/app/controllers/students_controller.rb
```diff
@@ -84,4 +87,10 @@ def destroy
   def find_student
     @student = Student.find(params[:id])
   end
+
+  def flash_errors
+    student.errors.full_messages.each do |error|
+      (flash[:error] ||= []) << error
+    end if student&.errors
```
jhuizy: You can chain the "safe navigation" operator which would remove the need for the trailing `if student&.errors`,

```suggestion
    student&.errors&.full_messages&.each do |error|
      (flash[:error] ||= []) << error
    end
```

I do think the chaining looks a bit ugly though. My personal preference is short circuiting (return early from the method) although this opinion may vary project to project

```ruby
return unless student&.errors
student.errors.full_messages.each do |error|
...
end

File: my-sms/app/views/students/index.html.erb [hunk 8]
```diff
@@ -1,35 +1,46 @@
...
+        <td><%=l student.created_at %></td>
+        <td>
+          <%= link_to 'Show', student, class: 'btn btn-default btn-xs' %>
+          <%= link_to t('.edit', :default => t("helpers.links.edit")),
+                      edit_student_path(student), :class => 'btn btn-default btn-xs' %>
+          <%= link_to t('.destroy', :default => t("helpers.links.destroy")),
+                      student_path(student),
+                      :method => :delete,
+                      :data => { :confirm => t('.confirm', :default => t("helpers.links.confirm", :default => 'Are you sure?')) },
+                      :class => 'btn btn-xs btn-danger' %>
```
jhuizy: Minor - but in the plus repo we prefer to use the `{ key: value }` syntax over the `{ :key => value }` syntax for hashes

my-sms/app/controllers/students_controller.rb
```diff
@@ -84,4 +87,10 @@ def destroy
   def find_student
     @student = Student.find(params[:id])
   end
+
+  def flash_errors
+    student.errors.full_messages.each do |error|
```
jhuizy: should this be `@student` instead?

```suggestion
    @student.errors.full_messages.each do |error|
```

my-sms/app/controllers/students_controller.rb
```diff
@@ -5,11 +5,14 @@ class StudentsController < ApplicationController
   decorates_assigned :student
 
   before_filter :find_student, only: %i[show edit update destroy]
+  before_render :flash_errors, only: %i[edit create update]
```
jhuizy, line 8: I'm thinking if you change the `student` to `@student` in the `flash_errors` method you can move this to `before_filter`. Great work on figuring out how to create your own callbacks, but I've not had to use custom callbacks like this before, so I'm a bit suspicious it's required here.

Also another opinion/personal note, I prefer calling the symbols directly rather than use the `%i` macro

```suggestion
  before_filter :flash_errors, only: [:edit, :create, :update]
```
```

Comment (josh-cooper) [reply my-sms/app/controllers/students_controller.rb:8]: The reason I had to create before_render is because rails does adds the error messages and the errors themselves to the model at different times. The messages are not added in until render is called. Not sure why, maybe because they want to maintain MVC separation?

File: my-sms/app/views/students/index.html.erb (hunk 8 above)
Comment (josh-cooper) [reply my-sms/app/views/students/index.html.erb]: Fixed in 5554e5f766237ceea0344d79b67ba1becda28022

File: my-sms/app/controllers/students_controller.rb (hunk 7 above)
Comment (josh-cooper) [reply my-sms/app/controllers/students_controller.rb]: Didn't know about the return unless syntax.
Switched to that in 640d900caff9530e476b4857a1e4842cc82e001d

File: my-sms/app/controllers/students_controller.rb (hunk 9 above)
Comment (josh-cooper) [reply my-sms/app/controllers/students_controller.rb]: Fixed in 19b82cd3dc7de83283b7660a0c09c3b44eb187c8

File: my-sms/app/controllers/students_controller.rb (hunk 7 above)
Comment (josh-cooper) [reply my-sms/app/controllers/students_controller.rb]: Also I added some specs for this section in 21acf9bd1752fa314e6f2e13f90c233c4e8a6543

File: my-sms/spec/controllers/students_controller_spec.rb [hunk 11]
```diff
@@ -125,6 +129,10 @@
         should be_success
         should render_template('edit')
       end
+
+      it "should propagate model errors to an error flash" do
+        expect { subject }.to change{ flash[:error] }.from(nil).to(["First name can't be blank"])
+      end
```
jhuizy, line 135: nice 👍

my-sms/app/controllers/students_controller.rb
jhuizy: The errors will be added when the model validations occur. These validations occur when the record is attempted to save to the database (eg. `@student.save`), or when you specifically call the `valid?` method (more info [here](https://guides.rubyonrails.org/active_record_validations.html)). 

Maybe try adding a `return unless @student.invalid?` line to your `flash_errors` method.

josh-cooper: That was a typo sorry 😅. What I meant was the error IDs themselves are added during validation, but corresponding messages for each type of error seem to be only added during render.

jhuizy: Hmm OK, let's leave for now.

Maybe in future, we can stick with original of showing errors above (using what you had before) and just indicate a simple message in the flash to indicate there are errors

something like

```ruby
def render_flash
  if @model.invalid?
    flash[:error] = "There are errors with your model"
  end
end
```

jhuizy (approved):

PR #7, opened 2021-02-10
Task 7 add title model

josh-cooper:
## Description
Create a Title model, populate it, and associate with Student model.
Assume this app is in production and there is data in the Students.title attr.

All tasks now complete.

## Breakdown of Commits
Add rake snapshot task : a74efac
 - Dump database in case everything goes irrevocably wrong
 
Migrate new title model : 415594a
 - Create new Title model
 - Create new titles table
 - Populate titles table with basic honorifics
 - Add Title association to Student model (has_one)
 - Update student table to reference Title model (title_id) and remove old title string column
 - Perform migration

Make tables name column non nullable : 14c14305da823123d7a32351cae62a6dcab11775

Refactor title migration : 453cd248af4c2a2ecca026d7bcae849771f6fa83
 - Unrecognised old titles in the students table are added to the titles table as part of the migration
 - If there are unrecognised titles in the students table warn instead of refusing to migrate
 - Run migration as two steps:
     - Create and populate titles table
     - Change Student title column from str to Title model
 - Find invalid titles via query interface
 - Access model properties via hash interface

Add title model association to Student : dfcd3ef2d13962d846d873b62f57f8db3744b3a8
 - Add one way belongs_to association
 - Override title getter to defer to title name unless specified

Update factories : 2bac8b9cdb43ef584aa5f341f3fadd5d7f89d74f
 - Add titles factory
 - Replace title on students factory to title association
 - Make factories migration friendly by only auto-loading factories when running specs

Fix full name spec to use Title association in factory : 0086c36ef1f1a88a0e987056de7804432141a52e

Add title model specs : fa498c484ce73c961075a7bcca20559da49e0d08

Remove leftover code that does nothing : dc0b638c58afe8b0c687e17b0e8882dd74c29a29

Keep old title column in migration : c5609c85cda148522e852927f980dc94c5ab2025
 - Extract title conversion to rake task

Refactor title spec : 9c1858a113f78d27f678467ebe5a7b6fb3efd5ed
 - Replace before filter with let bang method
 - Use expect syntax over should syntax

Refactor student decorate spec : fac18efd3fd5586517307c55737ab9d93f8b5dc9
 - Extract building of title association out into a let method

Change student title conversion rake task namespace : a13ee0af7704fc8eaa27b8d8702af1fedf76395a

Add dropdown for title selection : a96adfe1d74cb41cfc3a6f8044e513f952f70c31
 - Include collection select generated from Title collection
 - Add attr_accessor for title_id to Student model

Remove unnecessary before method : 91ddbb4af7744ba024098c91f9ab52348f7369f6

#### TODO Checklist
* [x] Migrate data from Student.title to the Title model. **\*\* in progress \*\***
  - [x] Perform migration
  - [x] Create and update specs
  - [x] Update model/controller code to work with Title model
  - Specs now succeed
* [x] Remove the Student.title attr
* [x] For the Student Edit page, add a dropdown to the title input box that will now pull from the Title model.

## Screenshot
image

### DIFF
--- a/my-sms/db/migrate/20210208041633_create_titles.rb
+++ b/my-sms/db/migrate/20210208041633_create_titles.rb
@@ -16,7 +16,7 @@ class CreateTitles < ActiveRecord::Migration
 
     # create titles table
     create_table :titles do |t|
-      t.string :name
+      t.string :name, null: false
 
       t.timestamps
     end
--- a/my-sms/db/migrate/20210208041633_create_titles.rb
+++ b/my-sms/db/migrate/20210208041633_create_titles.rb
@@ -1,0 +1,70 @@
+class CreateTitles < ActiveRecord::Migration
+  def up
+    available_titles = %w[Mr Mrs Miss Ms Master]
+
+    # do not perform migration if any unexpected titles found in Students
+    # may be more efficient to do via SQL query?
+    Student.find_in_batches do |batch|
+      batch.uniq{|s| s[:title]}.each do |s|
+        if available_titles.exclude? s.title
+          raise "Unexpected Student title #{s.title} in "\
+          "Student id #{s.id}. Migration expects titles "\
+          "in Students column to be one of #{available_titles}."
+        end
+      end
+    end
+
+    # create titles table
+    create_table :titles do |t|
+      t.string :name
+
+      t.timestamps
+    end
+
+    # pre-populate titles with english honorifics
+    available_titles.each do |name|
+      Title.create({ name: name })
+    end
+
+    # cache titles from db to avoid multiple queries
+    title_name_to_id = {}
+    Title.all.each do |title|
+      title_name_to_id[title.name] = title.id
+    end
+
+    # update students table
+    change_table :students do |t|
+      t.references :title
+    end
+
+    # add title model association
+    # without making assumptions about model
+    Student.reset_column_information
+    Student.find_each do |student|
+      student.update_attribute(:title_id, title_name_to_id.fetch(student.title, nil))
+      student.save
+    end
+
+    # remove old title (string) column
+    remove_column :students, :title
+  end
+
+  def down
+    # create temporary title_string column
+    add_column :students, :title_string, :string
+
+    # undo title model association without assuming title implementation in model
+    Student.find_each do |student|
+      student.update_attribute(:title_string, Title.find(student.title_id).name)
+    end
+
+    # remove title (model) column and rename title_string to title
+    change_table :students do |t|
+      t.remove :title_id
+      t.rename :title_string, :title 
+    end
+
+    # drop titles table
+    drop_table :titles
+  end
+end
--- a/my-sms/db/migrate/20210208041633_create_titles.rb
+++ b/my-sms/db/migrate/20210208041633_create_titles.rb
@@ -1,18 +1,18 @@
 class CreateTitles < ActiveRecord::Migration
-  def up
-    available_titles = %w[Mr Mrs Miss Ms Master]
+  HONORIFICS = %w[Mr Mrs Miss Ms Master]
 
-    # do not perform migration if any unexpected titles found in Students
-    # may be more efficient to do via SQL query?
-    Student.find_in_batches do |batch|
-      batch.uniq{|s| s[:title]}.each do |s|
-        if available_titles.exclude? s.title
-          raise "Unexpected Student title #{s.title} in "\
-          "Student id #{s.id}. Migration expects titles "\
-          "in Students column to be one of #{available_titles}."
-        end
-      end
+  def report_student_titles
+    invalid_titles = Student.where("title NOT IN (?)", HONORIFICS)
+    invalid_titles.each do |title|
+    logger.warn "Unexpected Student title #{s.title} in "\
+         "Student id #{s.id}. Migration expects titles "\
+         "in Students column to be one of #{HONORIFICS}."
     end
+  end
+
+  def up
+    # warn of any students with unexpected titles
+    report_student_titles
 
     # create titles table
     create_table :titles do |t|
@@ -22,48 +22,12 @@ class CreateTitles < ActiveRecord::Migration
     end
 
     # pre-populate titles with english honorifics
-    available_titles.each do |name|
+    HONORIFICS.each do |name|
       Title.create({ name: name })
     end
-
-    # cache titles from db to avoid multiple queries
-    title_name_to_id = {}
-    Title.all.each do |title|
-      title_name_to_id[title.name] = title.id
-    end
-
-    # update students table
-    change_table :students do |t|
-      t.references :title
-    end
-
-    # add title model association
-    # without making assumptions about model
-    Student.reset_column_information
-    Student.find_each do |student|
-      student.update_attribute(:title_id, title_name_to_id.fetch(student.title, nil))
-      student.save
-    end
-
-    # remove old title (string) column
-    remove_column :students, :title
   end
 
   def down
-    # create temporary title_string column
-    add_column :students, :title_string, :string
-
-    # undo title model association without assuming title implementation in model
-    Student.find_each do |student|
-      student.update_attribute(:title_string, Title.find(student.title_id).name)
-    end
-
-    # remove title (model) column and rename title_string to title
-    change_table :students do |t|
-      t.remove :title_id
-      t.rename :title_string, :title 
-    end
-
     # drop titles table
     drop_table :titles
   end
--- a/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
+++ b/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
@@ -1,11 +1,5 @@
 class MoveStudentTitleStrToTitleModel < ActiveRecord::Migration
   def up
-    # cache titles from db to avoid multiple queries
-    title_name_to_id = {}
-    Title.all.each do |title|
-      title_name_to_id[title.name] = title.id
-    end
-
     # update students table
     change_table :students do |t|
       t.rename(:title, :title_old)
--- a/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
+++ b/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
@@ -1,0 +1,44 @@
+class MoveStudentTitleStrToTitleModel < ActiveRecord::Migration
+  def up
+    # cache titles from db to avoid multiple queries
+    title_name_to_id = {}
+    Title.all.each do |title|
+      title_name_to_id[title.name] = title.id
+    end
+
+    # update students table
+    change_table :students do |t|
+      t.rename(:title, :title_old)
+      t.references :title
+    end
+
+    # add title model association
+    # without making assumptions about model
+    # add any titles not already in the db
+    Student.reset_column_information
+    Student.find_each do |student|
+      title = Title.where(name: student[:title_old]).first_or_create
+      student.update_attribute(:title_id, title.id)
+      student.save
+    end
+
+    # remove old title (string) column
+    remove_column :students, :title_old
+  end
+
+  def down
+    # re-create string title column
+    change_table :students do |t|
+      t.string :title_str
+    end
+
+    # undo title model association without assuming title implementation in model
+    Student.find_each do |student|
+      student.update_attribute(:title_str, Title.find(student[:title_id]).name)
+    end
+
+    # remove old title (model) column and rename title_str to title
+    remove_column :students, :title_id
+    rename_column :students, :title_str, :title
+  end
+end
--- a/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
+++ b/my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class MoveStudentTitleStrToTitleModel < ActiveRecord::Migration
   def up
     # update students table
@@ -6,18 +8,7 @@ class MoveStudentTitleStrToTitleModel < ActiveRecord::Migration
       t.references :title
     end
 
-    # add title model association
-    # without making assumptions about model
-    # add any titles not already in the db
-    Student.reset_column_information
-    Student.find_each do |student|
-      title = Title.where(name: student[:title_old]).first_or_create
-      student.update_attribute(:title_id, title.id)
-      student.save
-    end
-
-    # remove old title (string) column
-    remove_column :students, :title_old
+    logger.warn 'Run rake db:migrate:migrate_students to convert existing title Strings in title_old column.'
   end
 
   def down
--- a/my-sms/db/schema.rb
+++ b/my-sms/db/schema.rb
@@ -14,6 +14,7 @@
 ActiveRecord::Schema.define(:version => 20210212025754) do
 
   create_table "students", :force => true do |t|
+    t.string   "title_old"
     t.string   "first_name"
     t.string   "middle_name"
     t.string   "last_name"
--- a/my-sms/db/schema.rb
+++ b/my-sms/db/schema.rb
@@ -11,10 +11,9 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20210128041035) do
+ActiveRecord::Schema.define(:version => 20210208041633) do
 
   create_table "students", :force => true do |t|
-    t.string   "title"
     t.string   "first_name"
     t.string   "middle_name"
     t.string   "last_name"
@@ -23,6 +22,13 @@ ActiveRecord::Schema.define(:version => 20210128041035) do
     t.string   "gender"
     t.datetime "created_at",  :null => false
     t.datetime "updated_at",  :null => false
+    t.integer  "title_id"
+  end
+
+  create_table "titles", :force => true do |t|
+    t.string   "name"
+    t.datetime "created_at", :null => false
+    t.datetime "updated_at", :null => false
   end
 
 end
--- a/my-sms/db/schema.rb
+++ b/my-sms/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20210208041633) do
+ActiveRecord::Schema.define(:version => 20210212025754) do
 
   create_table "students", :force => true do |t|
     t.string   "first_name"
@@ -26,7 +26,7 @@ ActiveRecord::Schema.define(:version => 20210208041633) do
   end
 
   create_table "titles", :force => true do |t|
-    t.string   "name"
+    t.string   "name",       :null => false
     t.datetime "created_at", :null => false
     t.datetime "updated_at", :null => false
   end
--- a/my-sms/lib/tasks/migrate_student_title.rake
+++ b/my-sms/lib/tasks/migrate_student_title.rake
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+namespace :db do
+  namespace :migrate do
+    # add title model association
+    # without making assumptions about model
+    # add any titles not already in the db
+    desc 'Convert strings in title_old column to title model.'
+    task student_titles: :environment do
+      if Student.column_names.exclude? 'title_old'
+        raise 'Could not find title_old column in students table. Abandoning.'
+      end
+
+      Student.find_each do |student|
+        title = Title.where(name: student[:title_old]).first_or_create
+        student.update_attribute(:title_id, title.id)
+        student.save
+      end
+    end
+  end
+end
--- a/my-sms/lib/tasks/migrate_student_title.rake
+++ b/my-sms/lib/tasks/migrate_student_title.rake
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-namespace :db do
+namespace :data do
   namespace :migrate do
     # add title model association
     # without making assumptions about model
--- a/my-sms/spec/decorators/student_decorator_spec.rb
+++ b/my-sms/spec/decorators/student_decorator_spec.rb
@@ -6,7 +6,7 @@ describe StudentDecorator do
   it 'formats a full name with title, first name, middle name, last name' do
     student = build(
       :student,
-      title: 'Mr',
+      title: FactoryBot.build(:title, name: 'Mr'),
       first_name: 'John',
       middle_name: 'Adam',
       last_name: 'Smith'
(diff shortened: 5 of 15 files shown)


my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,70 @@
...
+    # may be more efficient to do via SQL query?
+    Student.find_in_batches do |batch|
+      batch.uniq{|s| s[:title]}.each do |s|
+        if available_titles.exclude? s.title
+          raise "Unexpected Student title #{s.title} in "\
+          "Student id #{s.id}. Migration expects titles "\
+          "in Students column to be one of #{available_titles}."
+        end
+      end
+    end
```
jhuizy: Why not create a title entry for each unique title found?

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,70 @@
...
+    # without making assumptions about model
+    Student.reset_column_information
+    Student.find_each do |student|
+      student.update_attribute(:title_id, title_name_to_id.fetch(student.title, nil))
+      student.save
+    end
+
+    # remove old title (string) column
+    remove_column :students, :title
+  end
```
jhuizy, line 28: Another approach (in pseudo-code) could be to:

- create `titles` table
- rename `title` to `title_old` in `students` table
- for each `student` in `students`: create a new title if it doesn't exist and set that as `student.title` (you can use `title.find_or_create_by(name: ...)`)
- remove `title_old` column in `students` table

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,70 @@
...
+    # update students table
+    change_table :students do |t|
+      t.references :title
+    end
+
+    # add title model association
+    # without making assumptions about model
+    Student.reset_column_information
+    Student.find_each do |student|
+      student.update_attribute(:title_id, title_name_to_id.fetch(student.title, nil))
```
jhuizy: As I was suggesting earlier, maybe something like 

```suggestion
      student.update(:title, Title.find_or_create_by_name(student.old_title))
```

my-sms/db/schema.rb
```diff
@@ -23,6 +22,13 @@
     t.string   "gender"
     t.datetime "created_at",  :null => false
     t.datetime "updated_at",  :null => false
+    t.integer  "title_id"
+  end
+
+  create_table "titles", :force => true do |t|
+    t.string   "name"
+    t.datetime "created_at", :null => false
+    t.datetime "updated_at", :null => false
```
jhuizy, line 32: Can rails 3.2 use `timestamps` for these fields?

my-sms/db/schema.rb
```diff
@@ -23,6 +22,13 @@
     t.string   "gender"
     t.datetime "created_at",  :null => false
     t.datetime "updated_at",  :null => false
+    t.integer  "title_id"
+  end
+
+  create_table "titles", :force => true do |t|
+    t.string   "name"
```
jhuizy: perhaps non null too?

my-sms/db/schema.rb
```diff
@@ -23,6 +22,13 @@
     t.string   "gender"
     t.datetime "created_at",  :null => false
     t.datetime "updated_at",  :null => false
+    t.integer  "title_id"
```
jhuizy, line 26: consider using `t.references`

my-sms/db/schema.rb
jhuizy: Ignore this, I'm looking at the wrong file

my-sms/db/schema.rb
jhuizy: Ignore this, I'm looking at the wrong file 😅

my-sms/db/schema.rb
josh-cooper: Seems like it, with some difficulty.
As far as I can tell Rail prefers to use datetime if available on the database. I defined t.timestamps in the migration file and haven't touched the schema.rb file.

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: I was wanting to remove as much model dependence as possible.
Update (I think) runs validations first which makes it dependent on the state of the Student model.

find_or_create_by_name looks good, wasn't aware of it. I was using the title_name_to_id hash to avoid searching the Titles column for each indivdual student's string but I prefer the simplicity of using find_or_create_by_name.

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: My thinking was that it would be better to be more explicit with what titles are available, since there is a small number.
Presumably the set of titles is a business decision since there is no documented functionality for users to define a custom title. This way that set of titles will be the same no matter where the migration is run.

my-sms/db/schema.rb
josh-cooper: I use this in the migration code, should I be editing the schema.rb too?

my-sms/db/schema.rb
josh-cooper: Fixed in 14c14305da823123d7a32351cae62a6dcab11775

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,70 @@
+class CreateTitles < ActiveRecord::Migration
+  def up
+    available_titles = %w[Mr Mrs Miss Ms Master]
+
+    # do not perform migration if any unexpected titles found in Students
+    # may be more efficient to do via SQL query?
+    Student.find_in_batches do |batch|
```
msxavi: Would this work instead? `Student.where("title NOT IN (?)", available_titles)`.

Ideally, we would not be doing this inside migrations or at least we would extract into a service class (`Migration::ReportStudentTitles`) so we can test it properly. And something like `Migration::UpdateStudentTitles` for what you do after.

Also, when you raise an error, unless you catch it elsewhere, the migration will stop. Not ideal here since creating titles table is independent of bad data. So you created a dependency which could be avoided.

There is quite a lot happening here which could have been done in separate steps:
1- Identifying/reporting on non-applicable or invalid student titles (independent process, no blockers);
2- Creating `titles` table and adding defaults (independent process, no blockers - decide what the defaults are);
3- Migrate student titles (dependent on `titles` table - here we would know what to do about invalid titles found in step one);

josh-cooper: > Would this work instead? Student.where("title NOT IN (?)", available_titles).

That's way better. I was trying to avoid SQL queries.
Should probably brush up on my SQL.

> Ideally, we would not be doing this inside migrations or at least we would extract into a service class (Migration::ReportStudentTitles) so we can test it properly. And something like Migration::UpdateStudentTitles for what you do after.

As in extract those out into their own classes then call them from the migrations?

> Also, when you raise an error, unless you catch it elsewhere, the migration will stop. Not ideal here since creating titles table is independent of bad data. So you created a dependency which could be avoided.

This was by design. Migrating the old title column to the new one is dependent on a set of titles, established before or during the migration. The migration does not run if this dependency has not been met as the decision of what to do with unexpected titles has not been made.
This is just me trying to assume the hypothetical business context around the SMS and not be destructive with the data, as these issues weren't addressed in the requirements. I'm happy to implement in whatever way is best.

> There is quite a lot happening here which could have been done in separate steps:
1- Identifying/reporting on non-applicable or invalid student titles (independent process, no blockers);
2- Creating titles table and adding defaults (independent process, no blockers - decide what the defaults are);
3- Migrate student titles (dependent on titles table - here we would know what to do about invalid titles found in step one);

Understood. To be clear, these should be separate migrations, and the identifying/reporting is its own process that can be run as a check before migration?

msxavi: @josh-cooper as you said, it is a design decision. I'm all for removing blockers instead of creating them. The current scenario is that your migration will fail if any record managed to go through with a title which is not in the list of `available_titles`. So my question is who has decided what `available_titles` should be and based on what requirement?

There were no validations to prevent invalid titles before and no data source for valid titles. That's why the first step would be running something like:
```
Student.select("DISTINCT(title)")
```
This would give you an idea of what "users" have been filling in and how many discrepancies there are between what the requirement defines as a valid title.

You would have to decide how your migration should handle these discrepancies, e.g: for titles that are not accepted as valid, leave student title blank or set some default value.

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: Ended up implementing something like this in addition to Xavier's comments in 453cd248af4c2a2ecca026d7bcae849771f6fa83
Works out much cleaner thanks.

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: Thanks I've refactored the migration into multiple steps and also incorporated some of Jordan's comments in 453cd248af4c2a2ecca026d7bcae849771f6fa83
Titles are now added if they are unrecognised in the second part of the migration.

msxavi (commented):
@josh-cooper Nice job. Although your tests are failing probably due to `title` refactor.

my-sms/db/migrate/20210208041633_create_titles.rb
jhuizy: Yea you're right, in a real world scenario there would probably be a list of valid titles given to you in the ticket, but we still probably don't want to raise an exception here. There are a few quick solutions I can think of to do:

- Don't raise an error but rather just log the titles and student ids that failed. That way we can create a script (rake task) in future to go through and fix those ones that failed
- Raise an error, but rollback any changes made so we're not left in an intermediate state
- Keep the old title in an existing column for a little while so that post migration we can easily see what titles still need to be migrated over without the end user noticing a change

josh-cooper: I ended up changing this part to just be a warning in 453cd24
I now just add unexpected titles to the table as you suggested.

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
```diff
@@ -0,0 +1,44 @@
+class MoveStudentTitleStrToTitleModel < ActiveRecord::Migration
+  def up
+    # cache titles from db to avoid multiple queries
+    title_name_to_id = {}
+    Title.all.each do |title|
+      title_name_to_id[title.name] = title.id
+    end
```
jhuizy: This doesn't appear to be used anywhere, can remove

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # add title model association
+    # without making assumptions about model
+    # add any titles not already in the db
+    Student.reset_column_information
+    Student.find_each do |student|
+      title = Title.where(name: student[:title_old]).first_or_create
+      student.update_attribute(:title_id, title.id)
+      student.save
+    end
```
jhuizy: This is looking good, maybe we should move this part here to a rake task. It can be tested separately and we can run it as many times as we need

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # remove old title (string) column
+    remove_column :students, :title_old
```
jhuizy: If you move the above to a rake task, we could leave this column in. What I've done before is left in columns for at least 1 release, and then removed after. Happy for others to chime in and see what they think though 👍

my-sms/spec/factories/students.rb
```diff
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :student do
-    title { %w[Mr Mrs Miss Ms].sample }
+    association :title, strategy: :build
```
jhuizy, line 5: Probably want the default to just be the create strategy. I think in this case it would be okay to leave as the default

```suggestion
    title
```

my-sms/spec/decorators/student_decorator_spec.rb
```diff
@@ -6,7 +6,7 @@
   it 'formats a full name with title, first name, middle name, last name' do
     student = build(
       :student,
-      title: 'Mr',
+      title: FactoryBot.build(:title, name: 'Mr'),
```
jhuizy: I would normally move the `title` here to a separate variable/let binding. Something like the following:

```ruby
subject { student.decorate }
let(:title) { create :title }
let(:student) { create :student, title: title, ... }

it "formats full name" do
  expect(subject.full_name).to eq "..."
end
```

my-sms/spec/models/title_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  context 'with valid params' do
+    it { should validate_presence_of(:name) }
+  end
+
+  context 'a title with the same name exists' do
+    let(:title) { build(:title, name: 'honorific') }
+    let(:other_title) { create(:title, name: 'honorific') }
+
+    before { other_title }
```
jhuizy: You can use the `bang` version of let, `let!` and then you won't need this before clause here

ie.
```ruby
let!(:other_title) { ... }

File: my-sms/spec/models/title_spec.rb [hunk 14]
```diff
@@ -0,0 +1,24 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Title, type: :model do
+  subject { title }
+  let(:title) { build(:title, title_params) }
+  let(:title_params) { {} }
+
+  context 'with valid params' do
+    it { should validate_presence_of(:name) }
```
jhuizy, line 11: Nice, I like the use of implicit subject here. One nitpick though is that they do prefer the `expect` matcher rather than `should` version, you can achieve the same like

```suggestion
    it { is_expected.to validate_presence_of(:name) }
```

my-sms/spec/models/title_spec.rb
jhuizy: This runs the `create(...)` straight away. The normal version of `let` is (as you have discovered) lazily loaded and only run once it's specifically called

jhuizy (commented):
Nice work, looking a lot better, just a few more things to move around and it will be good to go. Let me know if you have any issues or problems you want to discuss in regards to my comments 👍 I havn't been keeping track of other commenters so let me know if there's anything conflicting

my-sms/spec/factories/students.rb
josh-cooper: I use the build strategy because there's a standard set of titles created by the migration. This creates a conflict when creating the record because there is a uniqueness constraint on the title name.

I could replace this with something that won't appear in the database (e.g. title1, title2, title3,...) which would enable the create strategy to be used at the expense of readability.

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
jbat: @jhuizy @josh-cooper The pattern Jordan is referring to is common to what we do here at ReadyTech.
Another Ticket would be raised and added to a future sprint to drop the column.

The advantage is it acts like an extra safety mechanism.  Deleting Data (ie dropping columns) is highly risky and mitigating the risk where possible is the preferred direction.

If something was missed in the migration code above, and the data in title_old was removed straight away when the column was dropped `remove_column :students, :title_old` the process if getting the backup and restoring 300 hosts clients data in the Production environment is simply not an ideal situation to be in.

If the data was still in the title_old column and something went wrong, its a much cleaner/quicker/safer process to apply the fix.

my-sms/spec/decorators/student_decorator_spec.rb
josh-cooper: Fixed in fac18efd3fd5586517307c55737ab9d93f8b5dc9

my-sms/spec/models/title_spec.rb
josh-cooper: Fixed in 9c1858a113f78d27f678467ebe5a7b6fb3efd5ed

my-sms/spec/models/title_spec.rb
josh-cooper: Fixed in 9c1858a113f78d27f678467ebe5a7b6fb3efd5ed

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
josh-cooper: That makes a lot of sense. I've done both of these things in c5609c85cda148522e852927f980dc94c5ab2025

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
josh-cooper: I've implemented this as a part of c5609c85cda148522e852927f980dc94c5ab2025

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
josh-cooper: Fixed this in dc0b638c58afe8b0c687e17b0e8882dd74c29a29

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,34 @@
+class CreateTitles < ActiveRecord::Migration
+  HONORIFICS = %w[Mr Mrs Miss Ms Master]
```
jhuizy, line 2: Perhaps we can move this to the title model so it can be used elsewhere in the app

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def report_student_titles
+    invalid_titles = Student.where("title NOT IN (?)", HONORIFICS)
+    invalid_titles.each do |title|
+    logger.warn "Unexpected Student title #{s.title} in "\
+         "Student id #{s.id}. Migration expects titles "\
+         "in Students column to be one of #{HONORIFICS}."
```
jhuizy, line 9: Since we're no longer doing any data migrations here, perhaps we can move this method to the rake task

my-sms/db/migrate/20210208041633_create_titles.rb
```diff
@@ -0,0 +1,34 @@
...
+    create_table :titles do |t|
+      t.string :name, null: false
+
+      t.timestamps
+    end
+
+    # pre-populate titles with english honorifics
+    HONORIFICS.each do |name|
+      Title.create({ name: name })
+    end
```
jhuizy, line 27: This would then just become something like 

```ruby
Title.HONORIFICS.each do ...
```

my-sms/lib/tasks/migrate_student_title.rake
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+namespace :db do
```
jhuizy: I try avoid overwriting existing namespaces, maybe something like `:data` would be appropriate?

```
rake data:migrate:student_titles
```

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: I feel like that would create multiple sources of truth for what the available titles are. I'm happy to move it if you disagree.

my-sms/spec/factories/students.rb
jhuizy: In this case I don't think it will be an issue since we're no longer adding the data during the migration. Usually we would use a tool like [database_cleaner](https://github.com/DatabaseCleaner/database_cleaner) to clean the database inbetween test invocations anyways. You might need to just manually reset your test db otherwise for now just allow creation of dummy titles (like `test1`, `test2`, etc..)

my-sms/db/migrate/20210208041633_create_titles.rb
josh-cooper: The rake task just adds any existing titles to the new column. This is a warning for if the expected set of titles in `HONORIFICS` is significantly different to the ones in the old title column. Its not doing much anymore, so I can see why you would suggest removing it. I had just left it in because it did _something_ and was already there.
What do you think?

my-sms/lib/tasks/migrate_student_title.rake
josh-cooper: Fixed in a13ee0af7704fc8eaa27b8d8702af1fedf76395a

my-sms/db/migrate/20210208041633_create_titles.rb
jhuizy: Yea I get your thinking, if we move it to the rake task it should be OK still, you could even have it as a separate task, something like `report_invalid_titles`

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def down
+    # re-create string title column
+    change_table :students do |t|
+      t.string :title_str
+    end
+
+    # undo title model association without assuming title implementation in model
+    Student.find_each do |student|
+      student.update_attribute(:title_str, Title.find(student[:title_id]).name)
```
jhuizy, line 22: Nitpick: could simplify with `student.title.name`

jhuizy: You don't have to change this now, but it's worth knowing that you can also do `Student.includes(:title).find_each do ...` to avoid the n+1 query problem, but we might cover this a bit later

my-sms/db/migrate/20210208041633_create_titles.rb
jhuizy: Yes, you are correct - let's leave it here

my-sms/db/migrate/20210208041633_create_titles.rb
jhuizy: Actually, if we keep the `HONORIFICS` var in this file perhaps just leave this bit of code here

my-sms/spec/models/title_spec.rb
jhuizy: can now remove the `before {...}` line

jhuizy (commented):
Looking good, just a `before { .. }` that needs to be removd in the one of the tests and I will approve

my-sms/db/migrate/20210212025754_move_student_title_str_to_title_model.rb
josh-cooper: Was wanting to reduce model dependence by doing it this way.

josh-cooper:
> Looking good, just a `before { .. }` that needs to be removd in the one of the tests and I will approve

Oh right of course. Fixed in 91ddbb4af7744ba024098c91f9ab52348f7369f6

jhuizy (approved):

PR #8, opened 2021-02-16
Add faker

josh-cooper:
## Description
Add faker and use it for factory generated Student data.

## Breakdown of Commits
Add factory for generating demo data using faker : 7ed616d
- Add faker gem

#### TODO Checklist
* [x] Add faker gem

### DIFF
--- a/my-sms/Gemfile
+++ b/my-sms/Gemfile
@@ -15,7 +15,6 @@ gem 'twitter-bootstrap-rails'
 group :development do
   gem 'rubocop'
   gem 'spring-commands-rspec'
-  gem 'faker'
 end
 
 group :development, :test do
@@ -25,6 +24,7 @@ group :development, :test do
   gem 'reek'
   gem 'byebug'
   gem 'shoulda-matchers'
+  gem 'faker'
 end
 
 # Gems used only for assets and not required
--- a/my-sms/Gemfile
+++ b/my-sms/Gemfile
@@ -15,6 +15,7 @@ gem 'twitter-bootstrap-rails'
 group :development do
   gem 'rubocop'
   gem 'spring-commands-rspec'
+  gem 'faker'
 end
 
 group :development, :test do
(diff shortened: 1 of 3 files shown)


my-sms/Gemfile
```diff
@@ -15,6 +15,7 @@ gem 'twitter-bootstrap-rails'
 group :development do
   gem 'rubocop'
   gem 'spring-commands-rspec'
+  gem 'faker'
```
jbat: @josh-cooper looks like this should be in the next group .. ie to include :test as well as :development

my-sms/spec/factories/students.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'faker'
```
jbat: @josh-cooper I do not think you need this require at all.  try removing it.
if you do though .. best to add it to be available in all specs
https://github.com/josh-cooper/my-sms/blob/develop/my-sms/spec/rails_helper.rb

my-sms/Gemfile
josh-cooper: My preference was to only use it for development, for demo data. I was keeping a separate factory for tests using sequences as described in the Everyday Rails Guide. That way the test data is more predictable, making the tests more deterministic.
Though I'd prefer to follow the standard approach if that is to use it for tests as well.

my-sms/spec/factories/students.rb
josh-cooper: You're right I don't. Turns out I needed to move the calls to Faker into a block to make it work though.
Fixed in caed22a39dd60ef6306182bb45af87460a936a45

jbat (approved):

my-sms/Gemfile
josh-cooper: On second thought this might cause an error in the testing environment if any calls to Faker are left outside of blocks. Its safer just to include it as a test dependency even if it isn't used for testing at the moment.

Fixed in f1bf9a133f1e8d60d6bbead534339c5d48b6f8d1

PR #10, opened 2021-02-18
Task 9 create course model

josh-cooper:
## Description
Create a Course model with name, description, number_of_semesters, start_date, end_date, and allocation.
Allocation is the max number of students that can be assigned to the course.

## Breakdown of Commits
Create and migrate Course model : dff6536

Add course factories : f9549c2

Add Course model specs : ad9d650

Refactor date presence check in end_date_after_start validator : 5f3c7ddcae8e0aea443eb3eb0ab75e6ccb8868f0

#### TODO Checklist
* [x] Create a course model with the required attributes.


my-sms/app/models/course.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  validates :name, presence: true
+  validates :number_of_semesters, presence: true
+  validates :start_date, presence: true
+  validates :end_date, presence: true
+
+  validate :end_date_after_start
+
+  def end_date_after_start
+    return unless !end_date.blank? && !start_date.blank?
```
SyntheticDave: What do you think of flipping the condition? I find the following form of this check more readable:

```ruby
return if end_date.blank? || start_date.blank?
```

As an alternative that I also think is a little more readable (but like a little less):

```ruby
return unless end_date.present? && start_date.present?
```

SyntheticDave (commented):
Looks good @josh-cooper !
Just one suggestion.

my-sms/app/models/course.rb
josh-cooper: I like that better too. 
Fixed in 5f3c7ddcae8e0aea443eb3eb0ab75e6ccb8868f0

SyntheticDave (approved):
👍

PR #11, opened 2021-02-18
Task 10 create courses list

josh-cooper:
## Description
Create a List page for the Course model,  with pagination.
This is the first page for the Course model so the controller is created with the default CRUD routes. A partial is used to generalise the view behaviour of the Student model. Views for the other routes in the controller are left as the generator default splash pages.

## Breakdown of Commits
Add courses controller : 9506e30
 - Add controller with standard CRUD routes
 - Add course decorator
 - Add DEFAULT_PER_PAGE to Course model
 - Configure routing for courses resource

Extract students index page into partial : c62a123

Create course listing page : dcac4bc
 - Generate view stubs for each courses route
 - Use _crud_model partial to render courses list

Add courses controller specs : f768e77
 - Add before_filter for finding course from params
 - Automatically assign decoration in controller
 - Fix courses factory to output dates instead of datetimes

#### TODO Checklist
* [x] Create a List page for the Course model,  with pagination

  
image


my-sms/app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+  decorates_assigned :course
+
+  before_filter :find_course, only: %i[show edit update destroy]
+
+  # GET /courses
+  # GET /courses.json
+  def index
+    @courses = CourseDecorator.decorate_collection(
+      Course.paginate(page: params[:page], per_page: Course::DEFAULT_PER_PAGE)
+    )
```
msxavi, line 13: Do you still need `CourseDecorator.decorate_collection` after using `decorates_assigned :courses`? Or vice versa.

my-sms/app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    respond_to do |format|
+      format.html { redirect_to courses_url }
+      format.json { head :no_content }
+    end
+  end
+
+  def find_course
+    @course = Course.find(params[:id])
+  end
```
msxavi: By convention, we can have something that suits both find and initialise an instance:
```suggestion
  def load_course
    @course = if params[:id]
      Course.find(params[:id])
    else
      Course.new
    end
  end
```
So you can also use it before your `new` action:
```
before_filter :load_course, except: %i[index]
```

Optionally, you may want to tidy up your `new` and `show` actions, just like you did on `edit`.

my-sms/app/views/courses/index.html.haml
```diff
@@ -0,0 +1,3 @@
+- display_attrs = %i[id name number_of_semesters allocation start_date end_date]
+= render partial: 'shared/crud_models',
+  locals: {model_class: Course, model_collection: @courses, display_attrs: display_attrs}
```
msxavi, line 3: I like the idea, although you may think about using Presenters/Decorators (very similar patterns) to make your shared views more manageable and secure. For example, `display_attrs` could be extracted to Presenters. Check out this pattern: https://gist.github.com/somebox/5a7ebf56e3236372eec4#in-the-view

my-sms/app/controllers/courses_controller.rb
msxavi, line 88: By convention, we can have something that suits both find and initialise an instance:
```suggestion
  def load_course
    @course = if params[:id]
      Course.find(params[:id])
    else
      Course.new
    end
  end
```
So you can also use it before your `new` action:
```
before_filter :load_course, except: %i[index]
```

Optionally, you may want to tidy up your `new` and `show` actions, just like you did on `edit`.

my-sms/spec/controllers/courses_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+require 'rails_helper'
+
+RSpec.describe CoursesController, type: :controller do
+  let!(:course) { create(:course) }
+
+  describe 'GET #index' do
+    subject { get :index }
+    it 'returns a success response' do
+      should be_success
+      should render_template('index')
```
msxavi: Can you replace `should` with `expect`?

my-sms/app/controllers/courses_controller.rb
josh-cooper: Interesting convention. I've added it in caede4d9d3779522fffaaa6e565a74adae6cac83 .

my-sms/app/views/courses/index.html.haml
josh-cooper: That's interesting. Although I would prefer to continue using the partials pattern. 

The idea stated in the gist is to move business logic from controller and views. However, here the logic is purely view logic, which is traditionally separated into the View component. This also matches the `form_for` pattern logically (but not syntactically), in terms of what's inherited. 

I'm conscious of a rails trend to minimise code in views even if that code is view logic. Similar to desktop MVC where the view can directly observe the model and the 'view' part of the separation is preferred to act as templates rather than separating all view functionality. In that case, as you mention, I could move some logic into decorators that are already set up (such as `display_attrs`) but imo the separation is better here because the granularity is at the view level. For example, `display_attrs` would make also an appearance when creating the course edit/new pages but with different attrs.

my-sms/spec/controllers/courses_controller_spec.rb
josh-cooper: Fixed in 44d2330a93064812ea579fda2ed10b625b3cd1e9

my-sms/app/controllers/courses_controller.rb
josh-cooper: Fixed in 64eb1df70350c4c4ada0e6c1c3d509af8f5d6624

my-sms/app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+    respond_to do |format|
+      format.html # show.html.haml
+      format.json { render json: @course }
+    end
+  end
+
+  # GET /courses/new
+  # GET /courses/new.json
+  def new
+    @course = Course.new
```
msxavi, line 33: No longer required.

msxavi (commented):
One minor comment. Good job!

PR #12, opened 2021-02-19
Task 11 add course show edit

josh-cooper:
## Description
Add Show, Edit buttons / functionality to the Course List page.

## Breakdown of Commits
Add course show page : 2cdcf84a102dd9492caf2f5455c79df20daa3236
    
 - Extract student show method into partial
 - Use partial to render course show page

Create course form : 60526ff0b344b95ba36e0ac9f1ea3ac4dabea6ad

Add course edit page : 0beaf2a8d70df69124cd9abc3326d8b7f69bd9ff

Remove full width styling on number fields : 1d6a2ed78e913df541a15ed744df291fce6cc917
    
 - Add model_form stylesheet
 - Add css class to selectively unset width on form-controls

Add error flashes to courses : c82af02041525c01b19934614dbcc996313b27e1
    
 - Extract flash_errors method from students controller to application controller

Add courses error flashing spec : 901b383d638b5200f02745dd3df1d5e784def1bd

Add courses destroy specs : bab965ffde51d494676a00219ad0b8bf9b499e4e

Cleanup courses helper spec file : e603774df2d0bf781e5e0977290b896b1dcac801

#### TODO Checklist
* [x] Add show button/functionality
* [x] Add edit button/functionality

  
image


my-sms/app/views/courses/show.html.haml
```diff
@@ -1,2 +1,10 @@
-%h1 Courses#show
-%p Find me in app/views/courses/show.html.haml
\ No newline at end of file
+:ruby
+  display_attrs = %i[id name description number_of_semesters allocation start_date end_date]
+  display_ovrd = {
+    number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
+    start_date: { value: l(@course.start_date, format: :long) },
+    end_date: { value: l(@course.end_date, format: :long) }
+  }
```
jhuizy, line 7: Maybe these can be moved to the decorator?

my-sms/app/views/shared/_model_attr_list.html.haml
```diff
@@ -0,0 +1,15 @@
+- model_class = model_instance.class
+.page-header
+  %h1=t '.title', default: model_class.model_name.human.titleize
+%dl.dl-horizontal
+  - display_attrs.each do |e|
+    %dt.model-list-heading
+      %span.model-list-heading-text #{display_ovrd&.dig(e, :heading) || model_class.human_attribute_name(e)} :
+    %dd.model-list-value= display_ovrd&.dig(e, :value) || model_instance.send(e)
```
jhuizy, line 15: Hmm actually, I see what you're trying to do here... in this case I would probably prefer to be more verbose and explicit rather than being too tricky here. What might likely happen is you may have to display 1 field slightly different and you end up having to rewrite this anyways. Maybe just manually write out the fields?

my-sms/spec/controllers/courses_controller_spec.rb
```diff
@@ -116,6 +116,26 @@
...
+
+    # before { course }
+
+    it 'destroys the requested course' do
+      expect { subject }.to change(Course, :count).by(-1)
+    end
+
+    it 'redirects to the courses list' do
+      should be_redirect
+      should redirect_to(courses_url)
```
jhuizy: nitpick - use expect matchers (again, my bad for setting a bad example earlier on)

```suggestion
      is_expected.to be_redirect
      is_expected.to redirect_to(courses_url)
```

jhuizy (commented):
Looking good just some minor comments

my-sms/app/views/shared/_model_attr_list.html.haml
josh-cooper: That was the intention of the `display_ovrd` pattern. It turned out to be pretty powerful and saved a lot of duplication, so I would prefer to keep it in. E.g. I was able to render a list of courses during task 14 by injecting a partial: https://github.com/josh-cooper/my-sms/blob/task_14_create_institute_show_edit_pages/my-sms/app/views/institutes/show.html.haml

Xavier made a similar comment about using decorators in a presenter pattern in Task 10 which I didn't end up using for reasons documented here: https://github.com/josh-cooper/my-sms/pull/11#discussion_r578860041

jhuizy: It does save duplication but adds cognitive complexity and reduces flexibility, which is always the trade-off when DRY-ing code. IMO this saves only a few lines of code at best but makes it harder to maintain in future when we may have to add different fields and harder for others to reason about the code. Perhaps this could be helped if there were some comments discussing how this works (simple, 1-2 line comment would suffice) and maybe updating the variable name (what does `display_ovrd` mean?).

Also consider this scenario: The product owner now wants to display the number of semesters in a [badge](https://getbootstrap.com/docs/4.0/components/badge/), how would we implement this using this pattern? I know it's not always good to implement things using what-ifs but the lack of flexibility for future here is just unnecessary in comparison to the amount of code it saves 🙂

my-sms/spec/controllers/courses_controller_spec.rb
```diff
@@ -116,6 +116,26 @@
...
+        expect { subject }.to change { flash[:error] }.from(nil).to(['End date must be after start date'])
+      end
+    end
+  end
+
+  describe 'DELETE #destroy' do
+    let!(:course) { create(:course) }
+    subject { delete :destroy, { id: course.to_param } }
+
+    # before { course }
```
jhuizy, line 126: can ✂️ the commented code

my-sms/app/views/shared/_model_attr_list.html.haml
jhuizy: Happy to hear what you think though

my-sms/spec/controllers/courses_controller_spec.rb
josh-cooper: Fixed in 0a81a79d2c2fd14fc24a1a1e615b28555bb4fa17

my-sms/app/views/shared/_model_attr_list.html.haml
josh-cooper: I get your point about the tradeoff, but isn't the more important reason to DRY up code to abstract away repeated logic into a common interface so that there is a single source of truth, rather than save lines of code. I get that in a view this is largely acheived by css classes but I feel in this case the html description list structure contains enough logic to be worth DRYing up.

My preference to solve that scenario would be to write it as a partial and pass that to `display_ovrd` (display override). This plays nicely with Rails which seems to support moving display of model attributes into their own partials, evident by the inclusion of an `object` argument in the render partial function which only duplicates the `locals` functionality. It also DRYs up the display of that attribute, for instance if `number_of_semesters` was also displayed as a badge on the index page.

I've added some comments to the partial in f070161420f884ea5f5daa298bc53d171c85e649 
Although I'd still prefer to leave it in.

jhuizy: Yea I get your points, they all make sense, you can keep your code 😛 I just have 2 comments still to think about:

- If others need to use your code later on you need to make sure these kinds of smart things are well documented and easy to understand. Maybe it could be a well documented pattern for the entire project then it is OK, but if you were to introduce this to a large, existing codebase than perhaps there's a thought that yes, it saves you time and effort for creating and refactoring because it is a DRY approach (as well as being less fragile to logic changes), but is it worth it when I have to study the internals of the abstraction to figure out what's going on anyways? (and everyone else who needs to touch this code). 
- I still think there's a premature optimization going on here that feels like it would need to be hacked around for corner cases. I guess at that point in time you could refactor the whole thing since it's quite small anyways 🤷 

I do understand your points though, I have done this in projects countless times before (maybe where it did not make as much sense as here). The experiences were equally like:
- It saved the team time and effort in future work
- It didn't have an impact on future time and effort as opposed to the non-DRY approach
- It confused other or new developers
- It confused me looking at the code again for the first time in 6 months

This is obviously a really minor abstraction, but just something to think about 🙂

jhuizy (approved):

my-sms/app/views/shared/_model_attr_list.html.haml
josh-cooper: Thanks for the advice 👍 

```
I have done this in projects countless times before (maybe where it did not make as much sense as here). The experiences were equally like:

It saved the team time and effort in future work
It didn't have an impact on future time and effort as opposed to the non-DRY approach
It confused other or new developers
It confused me looking at the code again for the first time in 6 months
```
This is really helpful.

PR #14, opened 2021-02-23
Task 13 create institute list page

josh-cooper:
## Description
Create an Institute List Page displaying name, with pagination

## Breakdown of Commits
Add institutes controller : 8e1279b3bd7a98177f3f5e3ce3a9ae89536f2990
    
 - Add controller with standard CRUD routes
 - Add course decorator
 - Add DEFAULT_PER_PAGE to Course model
 - Configure routing for courses resource

Create institutes listing page : dfd7dd6e61619e9506fab714b04b9af8af44ad10
    
 - Use _crud_model partial to generate list

Add error flashes to institutes controller : 4ad9d39970e8116290b9ebb4a9141d0d06042b75

Add institutes controller specs : d76b18b3284b1207257ddff6d8b94194bc59bf2e

Replace should methods with expect : 7efe855bb57aa65e36a3a2068a938e4d525af88f

Remove redundant institute loading in controller : 214ee38a4f5d33fa49425ce75dd6c5f557300417

#### TODO Checklist
* [x] Institue list page
* [x] Pagination

  
Screen Shot 2021-02-23 at 5 32 04 pm

### DIFF
--- a/my-sms/app/assets/javascripts/institutes.js.coffee
+++ b/my-sms/app/assets/javascripts/institutes.js.coffee
@@ -1,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://jashkenas.github.com/coffee-script/
--- a/my-sms/app/assets/javascripts/institutes.js.coffee
+++ b/my-sms/app/assets/javascripts/institutes.js.coffee
@@ -1,3 +1,0 @@
-# Place all the behaviors and hooks related to the matching controller here.
-# All this logic will automatically be available in application.js.
-# You can use CoffeeScript in this file: http://jashkenas.github.com/coffee-script/
--- a/my-sms/app/assets/stylesheets/institutes.css.scss
+++ b/my-sms/app/assets/stylesheets/institutes.css.scss
@@ -1,0 +1,3 @@
+// Place all the styles related to the institutes controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
--- a/my-sms/app/assets/stylesheets/institutes.css.scss
+++ b/my-sms/app/assets/stylesheets/institutes.css.scss
@@ -1,3 +1,0 @@
-// Place all the styles related to the institutes controller here.
-// They will automatically be included in application.css.
-// You can use Sass (SCSS) here: http://sass-lang.com/
--- a/my-sms/app/controllers/institutes_controller.rb
+++ b/my-sms/app/controllers/institutes_controller.rb
@@ -1,9 +1,10 @@
-# frozen_string_literal: true
+# frozen_string_literal: false
 
 class InstitutesController < ApplicationController
   decorates_assigned :institute
 
   before_filter :load_institute, except: %i[index]
+  before_render :flash_errors, only: %i[edit create update]
 
   # GET /institutes
   # GET /institutes.json
@@ -88,5 +89,9 @@ class InstitutesController < ApplicationController
                  else
                    Institute.new(params[:institute])
     end
+  end
+
+  def flash_errors
+    super(@institute)
   end
 end
--- a/my-sms/app/controllers/institutes_controller.rb
+++ b/my-sms/app/controllers/institutes_controller.rb
@@ -1,0 +1,92 @@
+# frozen_string_literal: true
+
+class InstitutesController < ApplicationController
+  decorates_assigned :institute
+
+  before_filter :load_institute, except: %i[index]
+
+  # GET /institutes
+  # GET /institutes.json
+  def index
+    @institutes = InstituteDecorator.decorate_collection(
+      Institute.paginate(page: params[:page], per_page: Institute::DEFAULT_PER_PAGE)
+    )
+
+    respond_to do |format|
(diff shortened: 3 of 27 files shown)


my-sms/app/assets/javascripts/institutes.js.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://jashkenas.github.com/coffee-script/
```
jbat: File can be removed

my-sms/app/assets/stylesheets/institutes.css.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the institutes controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
jbat: File can be removed

my-sms/app/controllers/institutes_controller.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: false
+
+class InstitutesController < ApplicationController
+  decorates_assigned :institute
+
+  before_filter :load_institute, except: %i[index]
+  before_render :flash_errors, only: %i[edit create update]
```
jbat, line 7: Do you need flash_errors on `create` ?
What happens when you force an error on a new Institute ?

josh-cooper: Its a bit hard to test on institute as the actual edit/create/update functionality isn't included until the next task. I mainly included the `flash_errors` here to scaffold out the controller in preparation for that.

I did test it on student, which has the same functionality, and it works. `Create` redirects back to the new page if there's any errors on the model.
image

my-sms/app/controllers/institutes_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  # GET /institutes
+  # GET /institutes.json
```
jbat: Scaffold comments make it messy .. please remove

my-sms/app/models/institute.rb
```diff
@@ -4,4 +4,6 @@ class Institute < ActiveRecord::Base
   attr_accessible :name
 
   validates :name, presence: true
+
+  DEFAULT_PER_PAGE = 10
```
jbat: So this is the same value for Student and Coursel and most likely any new list page that will be added.
Points more towards one global constant.
How could you refactor this ?

my-sms/app/helpers/institutes_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+module InstitutesHelper
+end
```
jbat: Used for what exactly?
If you have a plan to use the file, can leave a TODO
`TODO: helper methods coming in next Institute feature for ... blah`

Otherwise its a waste of space and can be removed

my-sms/app/assets/javascripts/institutes.js.coffee
josh-cooper: Fixed in 67792896a31da920aff3f1f48a47966c3180d376

my-sms/app/assets/stylesheets/institutes.css.scss
josh-cooper: Fixed in 67792896a31da920aff3f1f48a47966c3180d376

my-sms/app/controllers/institutes_controller.rb
josh-cooper: Fixed in a3abd116c79511f50e413bc05c5b67319668f038

FYI I wasn't using scaffold generators for the institutes controller, I just duplicated them to match the original students controller.

my-sms/app/controllers/institutes_controller.rb
jbat: Interesting, so it fails and redirects back to the new method in the controller.
However the new method is not in `before_render :flash_errors, only:  %i[edit create update]`
How then is the flash_error showing for the new page ?
Fine to dive deeper on the next task .. when you have that edit functionality

my-sms/app/models/institute.rb
josh-cooper: I moved it to the application controller in 9e71a9ac8164c8919c65e335950a4b9e97092660 . Its additionally a better fit because the rest of the pagination is handled in the controller, and MVC wise it shouldn't really be part of the model. 

Decorators are arguably an even better fit, but I opted for the application controller as decorators are designed to act on model instances, whereas this is class level behaviour.

my-sms/app/helpers/institutes_helper.rb
josh-cooper: Removed in e5bb69888691df727165c5761d9a85cd7c3e351c

jbat (approved):
lg

my-sms/app/controllers/institutes_controller.rb
josh-cooper: My bad, its not technically a redirect. The controller just renders the `'new'` action. The before_render acts as soon as `render` is called so it doesn't matter which action gets rendered, just which route is called by the dispatcher.

PR #17, opened 2021-02-25
Task 15.2 add course and institute association

josh-cooper:
## Description
Partial completion of Task 15. Add association between Course and Insitute.
Institute has_many Courses. A Course belongs_to Institute.

## Breakdown of Commits
Migrate institute reference on courses table : c340ba9

Add association to course and institute models : a8b9362
- belogns_to on course
- has_many on institute

Update courses and institutes factories to reflect association : d504394

Add course institute association specs : b724d5a
- add attr access to associations on institute and course
- validate associated institute on courses

#### TODO Checklist
* [x]  Add relations, Institute has_many Courses. A Course belongs_to Institute.

### DIFF
--- a/my-sms/spec/models/course_spec.rb
+++ b/my-sms/spec/models/course_spec.rb
@@ -25,7 +25,7 @@ RSpec.describe Course, type: :model do
     it { is_expected.to be_valid }
   end
 
-  context 'with an invalid institute' do
+  context 'with an institute that has invalid params' do
     let(:institute) do
       institute = build(:institute, { name: nil })
       institute.save(validate: false)
--- a/my-sms/spec/models/course_spec.rb
+++ b/my-sms/spec/models/course_spec.rb
@@ -12,6 +12,7 @@ RSpec.describe Course, type: :model do
     it { is_expected.to validate_presence_of(:number_of_semesters) }
     it { is_expected.to validate_presence_of(:start_date) }
     it { is_expected.to validate_presence_of(:end_date) }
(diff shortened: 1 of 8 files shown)


my-sms/spec/models/course_spec.rb
```diff
@@ -23,4 +24,14 @@
...
   end
+
+  context 'with an invalid institute' do
+    let(:institute) do
+      institute = build(:institute, { name: nil })
+      institute.save(validate: false)
+      institute
+    end
+    let(:course_params) { { institute_id: institute.id } }
+    it { is_expected.to be_invalid }
```
jbat, line 35: @josh-cooper   I can see what your attempting here
It is a valid thought process.

Given that a course belongs_to an institute,
you are testing that a course with an empty institute is invalid.

I would argue that you are re-testing the belongs_to association
that has already been tested in the Active Record test in the Core Rails Rspec suite.
ie .. I dont see that this test adds any value.

You could, when the Controller functionality is added, cover this case there and confirm
that for an empty institute on a course, an error msg is raised for the user.

josh-cooper: I'm more trying to test that the model validations are also running on the association via `validates_associated`, which is a separate API to the belong_to association. An institute that on a course that exists but has no name should be invalidated.

I've changed the spec name to make my intention clearer in d062cf5c899fff55f42951681e8a5238af68ee7e

jbat (approved):

PR #18, opened 2021-03-01
Task 15.3 add course institute association views

josh-cooper:
## Description
Part 3 of Task 15. Task 15 is now complete.

Add gui features for the course and institute model association as follows:
- Add a dropdown to list available institutes on the course edit/create pages.
- Add institute name to course show page
- Display list of courses on institute show page
Also added a rake task to automatically generate demo data.

## Breakdown of Commits
Display institute name on course listing : 847a3c7
    
 - Overload to_s on institute to display name

Add edit dropdown to course form : a21934b

Add list of course links to institute show page : 0b84d78
    
 - Create _course partial

Add rake task for generating demo data : 8af01b8
    
 - Add option to create an associated institute on courses factory to prevent factory recursion
 - Use demo course factory for course association on demo institute factory
 - Change separate demo factories to a trait on the parent factory

#### TODO Checklist
* [x] Update Course Edit Page, add a new input field dropdown listing available Institutes 
* [x] Update Course List page to display Institute name 
* [x] Update the Institute List page, displaying all courses within an institute where the course name is the link to the Course Show page

  
image
image
image

### DIFF
--- a/my-sms/app/models/institute.rb
+++ b/my-sms/app/models/institute.rb
@@ -8,8 +8,4 @@ class Institute < ActiveRecord::Base
   validates :name, presence: true
 
   DEFAULT_PER_PAGE = 10
-
-  def to_s
-    name
-  end
 end
--- a/my-sms/app/models/institute.rb
+++ b/my-sms/app/models/institute.rb
@@ -8,4 +8,8 @@ class Institute < ActiveRecord::Base
   validates :name, presence: true
 
   DEFAULT_PER_PAGE = 10
+
+  def to_s
+    name
+  end
 end
--- a/my-sms/app/views/courses/show.html.haml
+++ b/my-sms/app/views/courses/show.html.haml
@@ -1,9 +1,9 @@
 :ruby
-  display_attrs = %i[id name description number_of_semesters allocation start_date end_date]
+  display_attrs = %i[id name description number_of_semesters allocation start_date end_date institute]
   display_ovrd = {
-    number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
-    start_date: { value: l(@course.start_date, format: :long) },
-    end_date: { value: l(@course.end_date, format: :long) }
+  number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
+  start_date: { value: l(@course.start_date, format: :long) },
+  end_date: { value: l(@course.end_date, format: :long) }
   }
 
 = render partial: 'shared/model_attr_list',
--- a/my-sms/app/views/courses/show.html.haml
+++ b/my-sms/app/views/courses/show.html.haml
@@ -1,9 +1,9 @@
 :ruby
   display_attrs = %i[id name description number_of_semesters allocation start_date end_date institute]
   display_ovrd = {
-  number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
-  start_date: { value: l(@course.start_date, format: :long) },
-  end_date: { value: l(@course.end_date, format: :long) }
+    number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
+    start_date: { value: l(@course.start_date, format: :long) },
+    end_date: { value: l(@course.end_date, format: :long) }
   }
 
 = render partial: 'shared/model_attr_list',
--- a/my-sms/app/views/institutes/show.html.haml
+++ b/my-sms/app/views/institutes/show.html.haml
@@ -1,3 +1,7 @@
--  display_attrs = %i[id name]
+:ruby
+  display_attrs = %i[id name courses]
+  display_ovrd = {
+  courses: { value: render(partial: 'course', collection: @institute.courses) }
+  }
 = render partial: 'shared/model_attr_list',
-  locals: {model_instance: @institute, display_attrs: display_attrs, display_ovrd: nil}
+  locals: {model_instance: @institute, display_attrs: display_attrs, display_ovrd: display_ovrd}
--- a/my-sms/lib/tasks/generate_demo_data.rake
+++ b/my-sms/lib/tasks/generate_demo_data.rake
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :data do
+  desc 'Generate demo data using FactoryBot.'
+  task generate: :environment do
+    require 'factory_bot_rails'
+    80.times { FactoryBot.create(:student, :demo) }
+    30.times { FactoryBot.create(:institute, :demo, with_courses: true) }
+  end
+end
--- a/my-sms/spec/factories/courses.rb
+++ b/my-sms/spec/factories/courses.rb
@@ -4,8 +4,6 @@ require 'faker'
 
 FactoryBot.define do
   factory :course do
-    transient { with_institute { true } }
-
     sequence(:name) { |n| "Course #{n}" }
     description 'This is a course description.'
     allocation { rand(1..50) }
@@ -19,7 +17,11 @@ FactoryBot.define do
       description { Faker::Lorem.paragraph }
       start_date Faker::Date.between(from: '2014-01-1', to: '2014-06-15')
       end_date Faker::Date.between(from: '2014-06-16', to: '2014-12-30')
-      institute { with_institute ? association(:institute_demo) : nil }
+      institute { association(:institute, :demo) }
+    end
+
+    trait :no_institute do
+      institute nil
     end
   end
 end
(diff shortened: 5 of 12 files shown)


my-sms/app/models/institute.rb
```diff
@@ -8,4 +8,8 @@ class Institute < ActiveRecord::Base
   validates :name, presence: true
 
   DEFAULT_PER_PAGE = 10
+
+  def to_s
+    name
+  end
```
jhuizy: can move to decorator

my-sms/lib/tasks/generate_demo_data.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :data do
+  desc 'Generate demo data using FactoryBot.'
+  task generate: :environment do
+    require 'factory_bot_rails'
+    80.times { FactoryBot.create(:student, :demo) }
+    30.times { FactoryBot.create(:institute, :demo, with_courses: true) }
```
jhuizy, line 8: Nice 👍 
Another way to do this is to use factory bot's inbuilt `create_list` method

```ruby
FactoryBot.create_list(:student, 30, :demo)

File: my-sms/app/views/courses/show.html.haml [hunk 3]
```diff
@@ -1,9 +1,9 @@
...
-  display_attrs = %i[id name description number_of_semesters allocation start_date end_date]
+  display_attrs = %i[id name description number_of_semesters allocation start_date end_date institute]
   display_ovrd = {
-    number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
-    start_date: { value: l(@course.start_date, format: :long) },
-    end_date: { value: l(@course.end_date, format: :long) }
+  number_of_semesters: { heading: Course.human_attribute_name('Semesters') },
+  start_date: { value: l(@course.start_date, format: :long) },
+  end_date: { value: l(@course.end_date, format: :long) }
   }
```
msxavi: What happened to your indentation here? How much more can this grow?

my-sms/spec/factories/courses.rb
```diff
@@ -12,12 +14,12 @@
...
     institute
 
-    factory :course_demo, class: :course do
+    trait :demo do
       name { Faker::Educator.course_name }
       description { Faker::Lorem.paragraph }
       start_date Faker::Date.between(from: '2014-01-1', to: '2014-06-15')
       end_date Faker::Date.between(from: '2014-06-16', to: '2014-12-30')
-      institute { association :institute_demo }
+      institute { with_institute ? association(:institute_demo) : nil }
```
jhuizy: Can also just override the `institute` field when creating one, or create a new trait.

eg.

```ruby
create(:course, :demo, institute: nil)
```

OR

```ruby
trait :no_institute do
  institute { nil }
end

...


create(:course, :demo, :no_institute)
```

jhuizy (commented):
Just move that method to a decorator and otherwise all good!

my-sms/app/views/institutes/show.html.haml
```diff
@@ -1,3 +1,7 @@
--  display_attrs = %i[id name]
+:ruby
+  display_attrs = %i[id name courses]
+  display_ovrd = {
+  courses: { value: render(partial: 'course', collection: @institute.courses) }
+  }
 = render partial: 'shared/model_attr_list',
-  locals: {model_instance: @institute, display_attrs: display_attrs, display_ovrd: nil}
+  locals: {model_instance: @institute, display_attrs: display_attrs, display_ovrd: display_ovrd}
```
msxavi, line 7: Since the last time I saw this, you're now passing another local variable and increasing complexity significantly. You now have to maintain two different variables which highly depend on each other `display_attrs` and  `display_ovrd`.

What is exactly the meaning of this?
```
courses: { value: render(partial: 'course', collection: @institute.courses) }
```
Are you really passing a full HTML inside a Hash? Why did you choose to override the values instead of assigning them from the beginning?

josh-cooper: This is actually a different partial to the one you are remembering I believe.

Overriding the values allows me to create a single source of truth for the layout structure, also making it easy to reproduce the same view logic for different models and weakly enforces some sort of visual consistency (via a kind of pavlovian response of convenience). `display_ovrd` achieves this at the minimal cost of flexibility by allowing the default title and content display to be overridden.

Passing partials into `display_ovrd` is by design as it allows complex view components to be rendered, however, in this case I could achieve the same result without a partial since the course list is pretty basic. The reason I'm using a partial here is to clean up the code here and provide some DRYness in case a list of courses needs to be rendered elsewhere. This is in line with the way Rails is intended to be used IMO, the guide describes using partials to render collections via the `collection` arg
https://guides.rubyonrails.org/v3.2.21/layouts_and_rendering.html#using-partials

There was already some discussion around this in a previous PR and it did have some benefit in later tasks so I would prefer to leave it in.

my-sms/app/views/courses/show.html.haml
josh-cooper: Not sure what happened there, fixed in a42eada4bd52a65646cd532b7e0afa0ed6387869

> How much more can this grow?

As in the contents of the variable on these lines?

my-sms/app/models/institute.rb
josh-cooper: Fixed in 4dac98b60e980e47b79aec7dbc42c7592b468c33

Also uncovered the issue that I hadn't set up decoration on course/institute association.

my-sms/spec/factories/courses.rb
josh-cooper: Thats way better thanks. Went for option 2.
Fixed in f5201bd4bc1bc0d6884387f28c006def12d6625e

my-sms/spec/factories/institutes.rb
```diff
@@ -6,9 +6,10 @@
...
     transient { with_courses { false } }
     sequence(:name) { |n| "Institute #{n}" }
-  end
-  trait :demo do
-    name { [Faker::Educator.university, Faker::Educator.secondary_school].sample }
-    courses { with_courses ? Array.new(rand(8)) { association(:course) } : [] }
+    courses { with_courses ? Array.new(rand(8)) { association(:course, :no_institute) } : [] }
+    trait :demo do
+      name { [Faker::Educator.university, Faker::Educator.secondary_school].sample }
+      courses { with_courses ? Array.new(rand(8)) { association(:course, :demo, :no_institute) } : [] }
```
jhuizy, line 12: Another tip: maybe just have a transient var for `courses_count`, and then you can just do

```ruby
courses { Array.new(courses_count) { assocation(:course, :demo, institutes: [instance]) } }
```

taken from the HABTM guide on factory bot: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#has_and_belongs_to_many-associations

jhuizy: AH, just realised it's not a HABTM., in that case, you can omit the `institutes: [instance]` from above

my-sms/app/views/institutes/show.html.haml
jhuizy: hmm, hang on, you're actually rendering the entire `list` view for coureses for each institute, yes? I think this will result in a lot of db queries (a la the n+1 problem), are you able to confirm the sql calls made when rendering this view?

Also I don't think it will look nice, we probably just want to show a list of course names for each `institute`, and maybe link to each course's show page?

my-sms/spec/factories/institutes.rb
josh-cooper: I saw that, just wanted to make it a random number of courses for more natural behavior. I thought about putting the `courses_count` inside `rand(n)` but it didn't really seem useful.

my-sms/app/views/institutes/show.html.haml
josh-cooper: Its just a partial displaying the course name as a link on the individual `institute` show page.
https://github.com/josh-cooper/my-sms/blob/f5201bd4bc1bc0d6884387f28c006def12d6625e/my-sms/app/views/institutes/_course.html.haml

There's no recursion here, rails handles the collection rendering efficientily when using the `collection` arg I believe. I can log the sql calls if you still want me to check.

jhuizy: oh cool 🙂 didn't see that partial there. Seems OK to me then as it continues from the pattern used previously

jhuizy (approved):

PR #20, opened 2021-03-03
Task 17 add note button to student

josh-cooper:
## Description
Add a button to Student show page to add a comment.
Add notes controller and configure as a nested resource on Student that will also work polymorphically on other models in future tasks.
Add notes controller specs.

## Breakdown of Commits
Remove colon in model_attr_list partial heading display : c81815f2c6d19a1bbc2f03828bcf891f654f23c6
    
 - Motivation was to prevent automatically adding colon when heading was overrided to increase flexibility
 - Decided to remove the colon alltogether because it looked cleaner

Add notes controller : 3a403993ba275fe4e646c975a6a7947c5a70590d
    
 - Load notable from params

Define notable resource on students : d3dd1935f55e0b70c14a97e52aaf2485dc5d285b
    
 - Add nested notable resource to students
 - Pass parent notable_type to nested note resource

Add notes list to students show page : 15f3e66051126048b9f637b112d52aa3dddaf184
    
 - Define partial for rendering a notes collection
 - Render that partial as part of the model_attr_list partial

Add create note button and form to model_attr_list : a8d0b929c5e5d514e7bd917d01b6d6ab4cbd32c7
    
 - Add shared note adding functionality to models
 - Assume for now that all models will have the ability to add notes
 - Can make add note button optional later if necessary
    This involves
 - Create new note form partial
 - Add new note button on `model_attr_list` that will display new note form
 - Refactor model_attr_list with comments to improve clarity

Remove unnecessary course creation statement : 6d55396116735e8910ee276b7fd083520b838dcf

Refactor shared crud model views to use a crud_button partial : 5b7878131f47b16f266c34404b9e1ce33737a45f
    
 - Only display buttons if route exists

Add notes index page : 5ba94f8a6c3711db53b6e0fda27fc45d987511d2
    
 - Add index method to controller
 - Refactor notable loading into its own method
 - Create index template with crud_models partial

Refactor note_resource route definition to allow additional args to be passed in : 3f893dafedd0170b8907c3dfb66ecb0f3b3fb365

Add notes show page : 06d0fe023d9b753c6d2d26a3ff3e246a530e20c8
    
 - Use model_attr_list partial
 - Check duck type before displaying add note button on model_attr_list
 - Add show route to notes resource

Try and redirect to notable explicitly on create and upate : bc378dd69aad11aa148f28c4fde05a7d0b6b7316
    
 - Fall back on redirecting back if notable is not available in params
 - Set frozen_string_literal false

Add notes controller specs : 54d0364d3d031b514c9e48ebe144b81af2669f2e

#### TODO Checklist
* [x] Add button to Student to add a note

  
image
image
image
image


my-sms/config/routes.rb
```diff
@@ -1,9 +1,22 @@
 # frozen_string_literal: true
 
 MySms::Application.routes.draw do
-  resources :students
+  get 'comments/create'
+
+  get 'comments/update'
+
+  get 'comments/destroy'
```
jhuizy: what are these routes for?

my-sms/config/routes.rb
```diff
@@ -1,9 +1,22 @@
...
+
+  def note_resource(**kwargs)
+    resources(:notes, notable_type: parent_resource.name.classify, only: %i[index create update destroy], **kwargs)
+  end
+
+  resources :students do
+    note_resource
+  end
```
jhuizy: why not just a nested notes resource?
https://guides.rubyonrails.org/routing.html#nested-resources

jhuizy (commented):
Couple of things;

- The task says to add a button to the student list page, are you able to do that?
- Maybe we can also show the note titles in the students list page? and add a link to each note that opens a modal and displays the full text? (or link to the note show page)

And possibly out of scope of this task
- Should there be a way to add/edit/delete the notes from the new/edit user pages?

jhuizy:
Maybe a simpler approach:

- Add `accepts_nested_attributes_for :notes` to the student model
- Edit the student new/edit forms to have a section for adding and deleting notes
- Perform note updates with the `Student.update` methods

my-sms/config/routes.rb
josh-cooper: Ah sorry missed those, removed them now.

my-sms/config/routes.rb
josh-cooper: This is a nested notes resource, I just add some extra params to the route because the note association is polymorphic and will not pass the student params as a notable.

The standard way to handle this problem in the Rails community is to add hidden fields to the form AFAICT but I thought it was better separation and more secure to handle it in the routing like this.

josh-cooper:
> Couple of things;
> 
> * The task says to add a button to the student list page, are you able to do that?
> * Maybe we can also show the note titles in the students list page? and add a link to each note that opens a modal and displays the full text? (or link to the note show page)
> 
> And possibly out of scope of this task
> 
> * Should there be a way to add/edit/delete the notes from the new/edit user pages?

I assumed that was an error and the task was meant to say to add the button to the show page. I forgot to chase up on that as I was working a few tasks ahead at the time.

I've added it in to the list page now anyway.

> Maybe a simpler approach:
> 
> * Add `accepts_nested_attributes_for :notes` to the student model
> * Edit the student new/edit forms to have a section for adding and deleting notes
> * Perform note updates with the `Student.update` methods

I didn't consider this. Its a lot better.

The way I have it implemented rn is more complex but it does have a slight advantage of being a little more polymorphic. As in I can CRUD notes through the one resource without having to maintain the functionality across every different controller (which for a note will hypothetically be a large number of models).

my-sms/config/routes.rb
jhuizy: Hmm I see...

What I would do if I wanted separate CRUD actions for the note is to do the following,

In routes, I would create custom routes for each notes section (could move this to a route concern- not sure if rails 3 supports this)

```ruby
resources :students do
  resources :notes, controller: "students/notes"
end
```

then create a controller `app/controllers/students/notes_controller.rb`, this will be a small file which should include a _controller concern_ for handing the note update and some private methods for finding the notable object. Something like:

```ruby
class Students::NotesController < ...
  include Concerns::NotesConcern

  private

  def find_notable
    @notable = Student.find(...) 
  end

  # probably not necessary if the notables all have the same name for the has_many
  def new_note
    @notable.notes.new
  end
end
```

Then I would move all the code from the notes controller (except the notable finding logic) into the concern placed in `app/controllers/concerns/notes_concern.rb`

jhuizy:
>The way I have it implemented rn is more complex but it does have a slight advantage of being a little more polymorphic. As in I can CRUD notes through the one resource without having to maintain the functionality across every different controller (which for a note will hypothetically be a large number of models).

This is what my last comment (on the code itself) addresses. Usually if you have shared logic between controllers you can extract that into a module/concern, and keep the non-shared logic (ie. finding the notable entity) in the specific controller.

josh-cooper:
> > The way I have it implemented rn is more complex but it does have a slight advantage of being a little more polymorphic. As in I can CRUD notes through the one resource without having to maintain the functionality across every different controller (which for a note will hypothetically be a large number of models).
> 
> This is what my last comment (on the code itself) addresses. Usually if you have shared logic between controllers you can extract that into a module/concern, and keep the non-shared logic (ie. finding the notable entity) in the specific controller.

I see, I'll change it to use `accepts_nested_attributes_for`. 
Fixed in e8b0e944759e9efb3c0154dbf6b57f16fdb09fd1 , although I've left the notes controller logic in for now because it would be useful if there was any necessary action to be routed through the Notes controller. Its not doing anything at the moment, however, so I can delete it if that's preferable.

my-sms/config/routes.rb
josh-cooper: This is not necessary anymore as I've moved to using `accepts_nested_attributes_for` to perform the note creation logic on the parent resource as you suggested.

That said, rather than separate CRUD actions for the note, what I was actually trying to achieve was to pass a request parameter with the `notable_type` when accessing the nested note resource. By default, Rails would only pass a `student_id` in the params which is fine if your nested resource is not polymorphic.

josh-cooper:
> > > The way I have it implemented rn is more complex but it does have a slight advantage of being a little more polymorphic. As in I can CRUD notes through the one resource without having to maintain the functionality across every different controller (which for a note will hypothetically be a large number of models).
> > 
> > 
> > This is what my last comment (on the code itself) addresses. Usually if you have shared logic between controllers you can extract that into a module/concern, and keep the non-shared logic (ie. finding the notable entity) in the specific controller.
> 
> I see, I'll change it to use `accepts_nested_attributes_for`.

my-sms/config/routes.rb
jhuizy: Ah, so you were trying to match all `{notable_type}/{id}/notes/...` routes to the notes controller? for all valid `notable_type` values?

I think in those cases, even though the model is polymorphic, I would still prefer to be explicit about the routes and controllers, and just use concerns or plain ruby modules to extract the shared logic.

josh-cooper: Yes but not by editing the actual routes. I was going to explicitly define the nested `note_resource` on each of Student, Course and Institute. All I was doing was passing in an extra request param of 'notable_type' so that the controller does not have to have a series of if/else statements to test if `student_id` exists, then `course_id`, etc.

my-sms/app/views/notes/_form_new.html.haml
```diff
@@ -1,11 +1,12 @@
...
-    %div.field
-        =f.label :title
-        %br
-        =f.text_field :title
-    %div.field
-        =f.label :content
-        %br
-        =f.text_field :content
+= form_for(notable) do |notable_f|
+    =notable_f.fields_for 'notes_attributes', notable.notes.build, index: nil do |f|
```
jhuizy, line 2: I've always used `new` instead of `build` and was wondering whether there was a difference...
turns out [as of rails 3.2.13](https://stackoverflow.com/a/12541282) they do the [same thing](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/relation.rb#L78) (for reference, the plus repo uses rails 3.2.22.5)

jhuizy:
>Its not doing anything at the moment, however, so I can delete it if that's preferable.

Yes please, if it's not doing anything no need to keep it around. In fact, over time if these things build up and can be dangerous as we don't know what functionality we can safely remove and how to test updates etc...

Also, do you need to update the student model and/or controller spec for the updating of notes?

my-sms/app/views/notes/_form_new.html.haml
josh-cooper: Ah good to know. I thought there was some sort of automatic behaviour `build` had when used in the form builder but guess not.

josh-cooper:
> > Its not doing anything at the moment, however, so I can delete it if that's preferable.
> 
> Yes please, if it's not doing anything no need to keep it around. In fact, over time if these things build up and can be dangerous as we don't know what functionality we can safely remove and how to test updates etc...
> 
> Also, do you need to update the student model and/or controller spec for the updating of notes?

Fixed in 9f2c36901ad755de6e97e9adbdf654321c38ad83 and added specs in 09889489fb29bdf07b7938a0ab6c73a833e7815d

Would you expect in the next two tasks, which are just adding the note button to the Course and Institute model, that I would need to duplicate the note updating specs from Student to the controller specs for those models too?

my-sms/app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+  def load_note
+    @note = if params[:id]
+              Note.find(params[:id])
+            else
+              note = Note.new(params[:note])
+              note[:notable_type] = @notable_type
+              note[:notable_id] = @notable_id
+              note
+    end
+  end
```
jhuizy, line 60: I still think to refactor most of this file into a concern, and put the notable loading logic into it's own controller. You'll be tripling up on the notes controller, but each controller will look as small as this:

```ruby
class Students::NotesController < ApplicationController
  include Concerns::NotesConcern

  private
  
  def load_notable
    @notable = Student.find(params[:student_id])
  end

end
```

And then in the concern you can remove the `load_notable` method (since that's implemented in each concrete controller) and then your `load_note` method becomes something like


```ruby
def load_note
  @note = @notable.notes.find_or_initialize_by(params[:id])
end
```

jhuizy (approved):

jhuizy:
> Would you expect in the next two tasks, which are just adding the note button to the Course and Institute model, that I would need to duplicate the note updating specs from Student to the controller specs for those models too?

Yes. Repeating stuff in specs is considered OK because we don't want to miss requirements

PR #21, opened 2021-03-12
Add cypress

josh-cooper:
## Description
Install and configure cypress. Add some acceptance testing for Task 4 based on the task spec.

## Breakdown of Commits
Install cypress: 495a1fa
- Install cypress via NPM
- Update dockerfile, docker-compose, and .devcontainer with cypress dependencies
- Update gitignore
- Add and install cypress-on-rails gem

Add acceptance tests for task 4: e07298a
- Create new display_student_list cypress spec file
- Fix div tag typo in will_paginate helper

Add github action for cypress tests: 263b9a5

#### TODO Checklist
* [x] Install and configure cypress
* [x] Add task 4 acceptance testing
* [ ] Add github action to run cypress tests


my-sms/spec/cypress/integration/acceptance/display_student_list.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      cy.get("tbody > tr").within(() => {
+        attrs.forEach((e, i) => cy.get("td").eq(i).contains(e));
+      });
+    });
+
+    it("Shows a student with a created at attribute in the correct format", () => {
+      cy.get("tbody > tr > td")
+        .eq(5)
+        .contains(/\d{4}-\d{2}-\d{2}\s\d{2}\:\d{2}\:\d{2}\s\w{3}/);
+    });
```
grantspeelman, line 57: - You want to be very careful about tying you tests to the markup, ie you dont want to have to update the test just because the markup changes but the content is the same. So I suggest you just assert that the Person's details are visible.
Also worth noting that these cypress tests will be much slower than unit tests, so I don't suggest doing a separate `it` for each assertion.

example
```suggestion
    it("lists the student", () => {
      // Shows a table with id, full name, email, birth date, gender, and created at columns
      cy.contains("Id").should("be.visible");
      cy.contains("Full name").should("be.visible");
      cy.contains("Email").should("be.visible");
      cy.contains("Birth date").should("be.visible");
      cy.contains("Gender").should("be.visible");
      cy.contains("Created at").should("be.visible");

      // it("Shows a student with the correct id, full name, email birth date, and gender attributes", () => {
      cy.contains("1").should("be.visible");
      cy.contains("John Smith").should("be.visible");
      cy.contains("john@example.com").should("be.visible");
      cy.contains("2011-04-09").should("be.visible");
      cy.contains("female").should("be.visible");
    });
```

grantspeelman (changes requested):
Nice Work so far 👍🏽 
Made a few suggestions.

